PR #26, opened 2011-07-26
Auto open specific downloads

cschomburg:
This adds a filter which opens specific downloads automatically. A signal `auto-open-filter` is called for each finished download and when it returns true, the download is opened via `open`.

Example usage:

``` lua
downloads.add_signal("auto-open-filter", function(destination, mime_type)
    return mime_type == "application/pdf"
end)
```

Currently the user has to define the applications for `open-file` manually, right? Maybe we could default to `xdg-open` in the configs, to serve as an example and fallback.

### DIFF
--- a/lib/downloads.lua
+++ b/lib/downloads.lua
@@ -93,18 +93,16 @@ status_timer:add_signal("timeout", function ()
     local running = 0
     for _, d in ipairs(downloads) do
         if is_running(d) then running = running + 1 end
-    end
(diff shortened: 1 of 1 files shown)


aidanholm:
Hi, and thanks for the pull request! Sorry it's been left for a long time!

I'd prefer this functionality to be implemented as a separate plugin, instead of directly modifying `downloads.lua`. I'd be happy to merge it with that change.

PR #27, opened 2011-07-31
Userscripts

karottenreibe:
- added auto-install feature when visiting `.user.js` URIs
- added confirmation dialog before the install


SolarAquarion:
Rebase or close?

SolarAquarion:
@aidanholm is this useful as in should i rebase it?

aidanholm:
Most of this is adding a wrapper for the GTK confirm dialog, which I would rather not do at the moment, just to keep the API surface smaller, and especially if there's only one module that would use the new API. 

@karottenreibe, thank you for your contribution, but for this reason I'd prefer not to merge this.

PR #32, opened 2011-08-23
Cookies

morgan-stephen:
Some updates to the cookies module related to session cookies. Session cookies will be cleared when the last window is closed and the libunique bindings are in use. Also provides some options to transform all cookies to session cookies, and to put an expiry time on session cookies.


aidanholm:
I'm sorry, but I'm afraid I'll have to close this PR unmerged :( WebKit no longer offers direct access to the cookie database, so the functionality added in this PR is not currently possible to implement through luakit.

PR #37, opened 2011-08-30
Class new

karottenreibe:
fixed luaH_class_new to use the table on the top of the stack instead of a fixed index (2)

also documented all the functions in common/luaclass.c, was heavily under-documented


mason-larobina:
Merged thanks!

PR #42, opened 2011-09-01
Added globals setting to allow/disallow opening of a URL in a new window.

jhamb:
Added globals setting to allow/disallow opening of a URL in a new window. 
Setting allow_new_windows to false in globals.lua should force all links to be opened in tabs instead of windows.


aidanholm:
Thanks for the contribution; I'm going to close this, however, as a configuration overhaul is going to be merged soon. I'll look into including such a setting after this has finished.

PR #114, opened 2012-08-02
URI expansion on Ctrl+Enter

nfbvieira:
This adds support for URI expansions on Ctrl+Enter for ```:open``` commands (including ```win``` and ```tab```), similarly to what happens in other browsers.

For example ```:open example<Ctrl+Enter>``` would expand to ```:open www.example.com```

On the other hand ```:open example abcd``` would maintain the default behavior (searching with the default search engine).


mason-larobina:
Merged, I shorted some var names & added :help text for this binding.

PR #146, opened 2012-10-18
Reset search results after "Clear results" click

dmedvinsky:
Say I do some history search, then click "clear results" button. After
that all the results are supposed to be gone. This blank white page
makes no use to the user anymore after that. So the right thing to do,
in my opinion, would be to reset the search term and show all the
history back again.


aidanholm:
Thanks @dmedvinsky for your contribution; I'm going to close this, however, because the history chrome JS will be rewritten soon, which will be an ideal time to reconsider how this should work.

PR #218, opened 2013-09-09
Adds option to save session on exit.

5paceManSpiff:
Does this through the save_on_exit bool in the global.lua config.


mason-larobina:
Thanks for the pull request.

This will be simple for people to patch in if they would like this to be the default close action. Currently this is handled by the 'ZZ' binding quite well.

PR #223, opened 2013-11-19
fix default useragent & add test

walpurgisriot:
Fixes unescaped wildcards in stock `globals.lua`, which cause part of Luakit versions like `0d5f4ab` to be improperly appended to the useragent, e.g.:
```
Mozilla/5.0 (Linux x86_64) AppleWebKit/537.6+ (KHTML, like Gecko) WebKitGTK+/1.10.2 luakit/0d5f4
```


mason-larobina:
Merged (without the tests, slightly overkill for fixing this regexp).

PR #243, opened 2014-06-02
Zoom level widget and basic auth storage

umonkey:
These are two extensions that I find useful for the general audience.  Basic auth can actually be treated as a bug fix for the login dialog which is not fully working out of the box.


SolarAquarion:
@aidanholm i guess this can be merged?

aidanholm:
I actually already rebased and merged the zoom level widget; I didn't get around to looking at the basic auth storage stuff yet though

aidanholm:
I'll close this for now, to keep the set of open PRs focused, as @umonkey appears inactive. If someone wants basic auth storage, this can be polished up then.

PR #265, opened 2015-01-19
Page step as global value

o-jasper:
Makes `page_step` a global value so people can easily set it to something else than `1.0` (like `0.9`)

*Trivial* change, if it ends up in there (obviously)dont mind neither way if you merge/manual edit.


mason-larobina:
Thanks, looks good.

o-jasper:
:+1:

PR #267, opened 2015-01-25
Separate view creation out from window creation

o-jasper:
Goal is to allow for multiple windows, i.e side-by-side. But needs a bunch more doing to get there;

* Plugins in existence typically want to talk to the current view. Some will want the list of views.

  Perhaps change `w.tabs` and `w.sbar` as needed, probably better to make `w.current_view()`/`w.view()` and change the necessary libs/plugins?
* The tabs highlight and change the current view, and perhaps a second kind if any of the views is showing it.
* Seems like some stuff, like `lib/download.lua` are actually reporting about something global? That doesnt quite fit well with the status-bar-per-view.


mason-larobina:
I don't think this is necessary. Multiple side-by-side windows can be provided by a tiling window manager.

Issue #169, opened 2013-01-14
Add support for SOCKS proxies


cf8:
Yes, i know there was this kind of request (#69) and it was closed.

But according to this news http://ftp.acc.umu.se/pub/gnome/sources/libsoup/2.39/libsoup-2.39.1.news libsoup can handle SOCKS proxies since version 2.39.1 and current version is 2.40.2.

AlmightyDeman:
Any news about socks5 support? This issue has been opened two years ago. Socks support would be great.
![photo_2015-04-26_03-15-59](https://cloud.githubusercontent.com/assets/2962598/7335464/b91f6834-ebc2-11e4-847c-0e3cb7ed7e3e.jpg)

aidanholm:
SOCKS support is working fine with updated WebKit2-based luakit; please reopen if there are more issues with it!

PR #268, opened 2015-01-30
Adds SOCKS support. (Yes, this uses remote DNS.)

Jookia:
I've tested this with SOCKS proxies and local/remote DNS, confirming soup passes DNS through the proxy.


jtacoma:
FYI, this also fixes #169.

mason-larobina:
Merged as https://github.com/luakit/luakit/commit/5c40f508d2881136aeae14129472a35a4d99b13b

mason-larobina:
(I might have forgotten how to use github, is there a way of marking that this was merged?)

PR #269, opened 2015-02-01
Asset from file, and applied to history chrome

o-jasper:
`lousy.load_asset` searches in the same places, as `require` does. Well, approximately.. I think instead of the current approach, it still needs to `?.lua` &rarr; `?.(...yourasset...)` and currently it just strips the `?.lua` and searches those directories. Why does lua deviate from the normal behavior of `PATH` anyway...

Also, history chrome can be configured a bit more fine-tunedly, by changing `globals.history.item_skelly`. The number of visits is defaultly included now. (`globals.config.search_select` tells what to select in the search, possibly to leave things out if you dont use them)

Can easily be applied to other chromes, but waiting on feedback on this one.


aidanholm:
Thanks for the PR! I notice there's already `lousy.load`; are there any situations where that couldn't be used? I like the idea of providing multiple resource trees, but I don't think that overloading `package.path` is a good idea; I'd prefer to have a separate (runtime-modifiable) set of directories. Ideally something simple like this:
```
lousy.load.dirs = { "some/path/", "/foo/bar", luakit.install_path }
```
With this change, there'd be no need for a new API.

aidanholm:
Unfortunately this PR has become rather outdated. While the core concept of splitting chrome pages into multiple files is worth consideration, the implementation would have to be wholly redone for luakit's new APIs and conventions. For this reason I'm closing this PR at this time.

PR #275, opened 2015-03-24
Fixed click() in follow.lua to trigger more events. 

ggVGc:
Triggering 'click' only doesn't work for some elements. This change makes follow.lua also trigger a sequence of 'mouseover', 'mousedown' and 'mouseup', making it work for a lot more clickable elements.

### DIFF
--- a/config/rc.lua
+++ b/config/rc.lua
@@ -116,6 +116,9 @@ require "follow"
 -- Match only hint labels
 --follow.pattern_maker = follow.pattern_styles.match_label
 
+-- Uncomment if you want to ignore case when matching
+--follow.ignore_case = true
+
 -- Add command history
 require "cmdhist"
 
--- a/lib/follow.lua
+++ b/lib/follow.lua
@@ -718,11 +718,17 @@ selectors = {
 
 evaluators = {
     click = [=[function (element) {
-        function click(element) {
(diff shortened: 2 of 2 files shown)


o-jasper:
mouseEvent("mouseover", ctrl);
    mouseEvent("mousedown", ctrl);
    mouseEvent("mouseup", ctrl);
    mouseEvent("click", ctrl);

Some chance that clicking twice this way does something unwanted somewhere, but i'd risk that.

ggVGc:
Yeah, good point, but I think it's fine. I've been using it for the past weeks, and not had any issues regarding clicking. I had some problems before, hence why I added this, and I stole the idea from vimb(another vim-like webkit-based browser), which I used for a while before and also didn't have problems there.

mason-larobina:
Hi @ggVGc thanks for the pull request.

The contents of 6d108aff013 look fine, the rest of the changes look unrelated.

I'll look into merging this soon, however I need to resurrect a way to build and run luakit on my laptop. Looks like the code has bit-rot a lot on archlinux (lua51, webkit1).

mason-larobina:
Merged 6d108af as https://github.com/luakit/luakit/commit/c05e751b63807dbb2b4c234433929c48cc6fd4a7

PR #280, opened 2015-04-27
Allow soup URIs that are invalid for HTTP

### LINKED ISSUE #169: Add support for SOCKS proxies

jtacoma:
This fixes #169 and "adds" support for SOCKS proxies by removing the
requirement that a URI property set on a GObject must be valid for HTTP.

### DIFF
--- a/common/property.c
+++ b/common/property.c
@@ -102,7 +102,7 @@ luaH_gobject_set(lua_State *L, property_t *p, gint vidx, GObject *object)
         else
             tmp.c = g_strdup_printf("http://%s", tmp.c);
         u = soup_uri_new(tmp.c);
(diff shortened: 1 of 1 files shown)


jtacoma:
It's pretty easy to imagine that SOUP_URI_VALID_FOR_HTTP was not an optional check, and that this change will therefore cause some other problems.  However, I've tested with invalid URIs and the results seem ok (we still see a sensible error message).  I also configured a socks5:// proxy URI and it worked as expected.

jtacoma:
Never mind this, see #268 instead.

PR #361, opened 2017-07-24
Fix for compatability with urxvt

gleachkr:
At least in xterm and urxvt, the -e flag is supposed to be followed by a command and then a series of arguments to that command. The quotes here (now removed) have the effect of forcing urxvt to read the entire string as the command, which will cause it to fail when a file or line is specified.

### DIFF
--- a/lib/editor.lua
+++ b/lib/editor.lua
@@ -6,7 +6,6 @@
 --
 -- @module editor
 
-local globals = require("globals")
 local capi = { luakit = luakit }
 
 local _M = {}
--- a/lib/editor.lua
+++ b/lib/editor.lua
@@ -20,8 +20,6 @@
 -- @copyright 2017 Graham Leach-Krouse
 -- @copyright 2017 Aidan Holm
 
-local capi = { luakit = luakit }
-
 local _M = {}
 
 -- substitute in common values from the environment.
@@ -71,7 +69,7 @@ _M.edit = function (file, line)
         line = line or 1,
     }
     local cmd = string.gsub(_M.editor_cmd, "{(%w+)}", subs)
-    capi.luakit.spawn(cmd)
+    luakit.spawn(cmd)
 end
 
 return _M
--- a/lib/editor.lua
+++ b/lib/editor.lua
@@ -1,32 +1,76 @@
 --- Text editor launching functionality.
 --
 -- This module is primarily for use by other Lua modules that wish to
--- allow the user to edit a particular text file. This module does not
--- have any user-facing modifications or features.
+-- allow the user to edit a particular text file. The default is to guess at the
+-- shell command to open a text editor from environment variables. To override
+-- the guess, replace `editor.cmd_string`. This can be done manually, as follows:
+--
+--     local editor = require "editor"
+--     editor.editor_cmd = "urxvt -e nvim {file} +{line}"
+--
+-- Before running the command, `{file}` will be replaced by the name of the file
+-- to be edited, and `{line}` will be replaced by the number of the line at
+-- which to begin editing. This module also supplies several builtin command
+-- strings, which can be used like this:
+--
+--     local editor = require "editor"
+--     editor.editor_cmd = editor.builtin.urxvt
 --
 -- @module editor
+-- @copyright 2017 Graham Leach-Krouse
+-- @copyright 2017 Aidan Holm
 
 local capi = { luakit = luakit }
 
 local _M = {}
 
+-- substitute in common values from the environment.
+local env_sub = function (s)
+    local subs = {
+        term =  os.getenv("TERMINAL") or "xterm",
+        editor = os.getenv("EDITOR") or "vim"
+    }
+    return string.gsub(s,"{(%w+)}", subs)
+end
+
+--- Built in substitution strings. Includes
+--
+-- * `default` (attempts to extract a terminal and editor from environment
+-- variables, and otherwise falls back to xterm and vim)
+-- * `xterm`
+-- * `urxvt`
+-- * `urxvtc`
+-- * `xdg_open`
+--
+-- @type table
+-- @readonly
+_M.builtin = {
+    default = env_sub("{term} -e '{editor} {file} +{line}'"),
+    xterm = env_sub("xterm -e {editor} {file} +{line}"),
+    urxvt = env_sub("urxvt -e {editor} {file} +{line}"),
+    urxvtc = env_sub("urxvtc -e {editor} {file} +{line}"),
+    xdg_open = env_sub("xdg-open {file}")
+}
+
+--- The shell command used to open the editor.
+--
+-- @type string
+-- @readwrite
+_M.editor_cmd = _M.builtin.default
+
 --- Edit a file in a terminal editor in a new window.
 --
 -- * Can't yet handle files with special characters in their name.
--- * Can't yet use a graphical text editor (terminal only).
 -- * Can't determine when text editor is closed.
 --
 -- @tparam string file The path of the file to edit.
--- @tparam number line The line number at which to begin editing.
+-- @tparam[opt] number line The line number at which to begin editing.
 _M.edit = function (file, line)
     local subs = {
-        term = globals.term or os.getenv("TERMINAL") or "xterm",
-        editor = globals.editor or os.getenv("EDITOR") or "vim",
         file = file,
-        line = line and " +" .. tostring(line) or "",
+        line = line or 1,
     }
-    local cmd_tmpl = "{term} -e '{editor} {file}{line}'"
-    local cmd = string.gsub(cmd_tmpl, "{(%w+)}", subs)
+    local cmd = string.gsub(_M.editor_cmd, "{(%w+)}", subs)
     capi.luakit.spawn(cmd)
 end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 ### Fixed
 
  - Fixed code-blocks in documentation being formatted incorrectly.
+ - Fixed incompatibility of `editor.lua` with urxvt.
+
+### Changed
+
+ - `editor.lua` now uses substitution strings, rather than `global` to determine which editor to open.
 
 ## [2017-07-26]
 


aidanholm:
Unfortunately this breaks compatibility with my terminal! Terminals and editors (command-line or not) all accept differently-named arguments. I think a better fix would be to use a format string like pentadactyl's ["editor" option](http://5digits.org/help/pentadactyl/options.xhtml#%27editor%27), plus perhaps a few predefined ones. That way the user can do something like:
```lua
editor.editor_command = editor.builtin.vim_urxvt
```

gleachkr:
Oops, sorry about that. Format strings sounds like a good idea. I can give it a try if you'd like.

Would the idea be to add `editor_command` to `globals`? 

If so, would that deprecate `globals.editor`, or would it be better to retain the old behavior as a default (perhaps building in a few predefined special cases, depending on `globals.editor`)?

aidanholm:
No problem; that'd be great! Should probably also take a look at how vimium and other systems implement editor opening.

I think `editor_command` would be better on the `editor` module; `globals` is currently a bit cluttered. Not sure about `globals.editor` yet; perhaps just a warning if it's set to anything at all, so people know to switch?

gleachkr:
Here's a first pass. There's no warning yet for `globals.editor`, but does this look otherwise roughly on track?

lib/editor.lua
```diff
@@ -6,11 +6,41 @@
...
+        term =  os.getenv("TERMINAL") or "xterm",
+        editor = os.getenv("EDITOR") or "vim"
+    }
+    return string.gsub(s,"{(%w+)}", subs)
+end
+
+local builtin = {
+    default = env_sub("{term} -e '{editor} {file}{line}'"), 
+    xterm = env_sub("xterm -e {editor} {file}{line}"), 
+    urxvt = env_sub("urxvt -e {editor} {file}{line}"),
```
aidanholm: The concern here is that different editors have different and incompatible ways to specify the line number.

lib/editor.lua
```diff
@@ -6,11 +6,41 @@
...
+_M.editor_cmd = builtin.default
+
+-- Built in substitution strings. includes
+--
+-- * default (attempts to term and editor from environment variables, and
+-- otherwise falls back to xterm and vim)
+-- * xterm
+-- * urxvt
+-- * urxvtc
+_M.builtin = builtin
```
aidanholm: Should just declare the table here, rather than separately as a local table as well.

lib/editor.lua
```diff
@@ -21,13 +51,10 @@ local _M = {}
 -- @tparam number line The line number at which to begin editing.
 _M.edit = function (file, line)
     local subs = {
-        term = globals.term or os.getenv("TERMINAL") or "xterm",
-        editor = globals.editor or os.getenv("EDITOR") or "vim",
         file = file,
         line = line and " +" .. tostring(line) or "",
```
aidanholm: Would probably be a good idea to change this to `line = line or 1` and move the " +" to the format string.

aidanholm:
Mostly good. Theres a few minor changes to make. 

An alternative to a set of default format strings that occurred to me was to use two format strings internally: one to build the editor command, and another to wrap that with the command needed to start the terminal, switching between the latter based on $TERM. It's a little more complicated, though, and quoting could be an issue, so we should probably stick with the current approach for now.

lib/editor.lua
```diff
@@ -6,11 +6,41 @@
...
+end
+
+local builtin = {
+    default = env_sub("{term} -e '{editor} {file}{line}'"), 
+    xterm = env_sub("xterm -e {editor} {file}{line}"), 
+    urxvt = env_sub("urxvt -e {editor} {file}{line}"),
+    urxvtc = env_sub("urxvtc -e {editor} {file}{line}")
+}
+
+-- the default editor command, which can be overridden by a string like, for
```
aidanholm: Documentation comments start with three dashes not two; same in other places. You'll need to add some annotations like `@readonly` otherwise `make apidoc` will complain; feel free to ask about errors that come up, as I never tidied it up much.

lib/editor.lua
```diff
@@ -6,11 +6,41 @@
...
+-- example, 'editor.editor_cmd = urxvt -e nvim {file}{line}', where '{file}'
+-- gives a file name, and '{line}' gives the line number. Alternatively, you can
+-- use a builtin, setting 'editor.editor_cmd = editor.builtin.urxvt'
+_M.editor_cmd = builtin.default
+
+-- Built in substitution strings. includes
+--
+-- * default (attempts to term and editor from environment variables, and
```
aidanholm: Three dashes here as well

lib/editor.lua
```diff
@@ -6,11 +6,41 @@
...
+    urxvtc = env_sub("urxvtc -e {editor} {file}{line}")
+}
+
+-- the default editor command, which can be overridden by a string like, for
+-- example, 'editor.editor_cmd = urxvt -e nvim {file}{line}', where '{file}'
+-- gives a file name, and '{line}' gives the line number. Alternatively, you can
+-- use a builtin, setting 'editor.editor_cmd = editor.builtin.urxvt'
+_M.editor_cmd = builtin.default
+
+-- Built in substitution strings. includes
```
aidanholm: Three dashes here as well

gleachkr:
Ok, this covers allowing custom formatting for line numbers in the `editor_cmd` string, and fixes up the comments.

aidanholm:
- Remember to run `make run-tests`; I'm getting the following issues:
```
FAIL style/test_luacheck / test_luacheck
  tests/style/test_luacheck.lua:102: Luacheck messages:
    lib/editor.lua:15: (19:21):           line contains trailing whitespace
    lib/editor.lua:32: (62:62):           line contains trailing whitespace
    lib/editor.lua:33: (57:57):           line contains trailing whitespace
    lib/editor.lua:39: (12:12):           trailing whitespace in a comment
    lib/editor.lua:45: (49:49):           trailing whitespace in a comment
    lib/editor.lua:46: (3:3):             trailing whitespace in a comment
    lib/editor.lua:48: (3:3):             trailing whitespace in a comment
    lib/editor.lua:64: (5:15):            setting non-standard global variable 'linepattern'
    lib/editor.lua:64: (77:77):           line contains trailing whitespace
    lib/editor.lua:67: (37:47):           accessing undefined variable 'linepattern'
    lib/editor.lua:67: (69:69):           line contains trailing whitespace
```

- Please rebase your commits onto the latest `develop`, and add an entry to the changelog.

lib/editor.lua
```diff
@@ -6,11 +6,50 @@
...
+-- @type table
+-- @readonly
+_M.builtin = {
+    default = env_sub("{term} -e '{editor} {file} {+line}'"), 
+    xterm = env_sub("xterm -e {editor} {file} {+line}"), 
+    urxvt = env_sub("urxvt -e {editor} {file} {+line}"),
+    urxvtc = env_sub("urxvtc -e {editor} {file} {+line}")
+}
+
+--- the default editor command, which can be overridden by a string like, for
```
aidanholm: Change to "The shell command used to open the text editor."

No need to mention that it can be overriden, that's what the `@readwrite` means :)

But please move the example (which is good!) to the top comment of the file, and remove the now-incorrect text "This module does not have any user-facing modifications or features." See `lib/follow.lua` for examples on formatting this.

lib/editor.lua
```diff
@@ -6,11 +6,50 @@
...
+-- example, 
+--
+-- `editor.editor_cmd = urxvt -e nvim {file}{line}`
```
aidanholm: Remember to surround the command in quotes:
```
`editor.editor_cmd = " .... "`
```
And once you move this to the top, indent it by four spaces instead of using backticks, to make it a large code block

lib/editor.lua
```diff
@@ -6,11 +6,50 @@
...
+--
+-- where `{file}` gives a file name, and `{line}` gives the line number.
+--
+-- Alternatively, you can use a builtin, setting 
+-- 
+-- `editor.editor_cmd = editor.builtin.urxvt`
```
aidanholm: Ditto for the indentation here

lib/editor.lua
```diff
@@ -6,11 +6,50 @@
 --
 -- @module editor
```
aidanholm: Please add both yourself and me (I forgot before!) as authors and copyright to the header

lib/editor.lua
```diff
@@ -20,14 +59,19 @@ local _M = {}
...
 _M.edit = function (file, line)
+    --extracts the pattern from the editor command into which we will substitute
+    --the line number, if a line number is supplied.
+    linepattern = string.gsub(_M.editor_cmd, ".*{([^}]*line[^}]*)}.*", "%1") 
     local subs = {
-        term = globals.term or os.getenv("TERMINAL") or "xterm",
-        editor = globals.editor or os.getenv("EDITOR") or "vim",
         file = file,
-        line = line and " +" .. tostring(line) or "",
+        line = line and string.gsub(linepattern,"line", line) or "",
```
aidanholm: This is very clever, but not what I had in mind :) My preference is to write the format string as, e.g:
```lua
"{term} -e {editor} {file} +{line}"
```
Then, if the lua caller didn't specify a line, we just use line 1 instead, and the result is the same. This will result in code that is a lot easier to read.

lib/editor.lua
gleachkr: For some reason, this bit of syntax (code block via a four-space indentation) isn't working for me. 

I notice that in the documentation for follow.lua (generated by my system luakit, rather than the development one that I'm working on) the four-space indented sections are garbled. Here's a screenshot:

2017-07-25-144240_685x206_scrot

Any ideas?

gleachkr:
I've tried to squash these properly. I can squash further if that would be preferable.

I'm able to run the tests (thanks for the tip about these), and there are now no errors involving `editor.lua`.

If this looks good to you, and if we can figure out the code-block issue, then I'll go ahead and rebase on the latest development branch, and update the changelog.

lib/editor.lua
```diff
@@ -1,16 +1,60 @@
...
+-- * `urxvt`
+-- * `urxvtc`
+--
+-- @type table
+-- @readonly
+_M.builtin = {
+    default = env_sub("{term} -e '{editor} {file} +{line}'"),
+    xterm = env_sub("xterm -e {editor} {file} +{line}"),
+    urxvt = env_sub("urxvt -e {editor} {file} +{line}"),
+    urxvtc = env_sub("urxvtc -e {editor} {file} +{line}")
```
aidanholm: Might be a good idea to add an xdg-open one here as well

lib/editor.lua
```diff
@@ -1,16 +1,60 @@
...
--- allow the user to edit a particular text file. This module does not
--- have any user-facing modifications or features.
+-- allow the user to edit a particular text file. The default is to guess at the
+-- shell command to open a text editor from environment variables. To override
+-- the guess, replace `editor.cmd_string`. This can be done manually, as follows:
+--
+--     editor = require "editor"
+--     editor.editor_cmd = "urxvt -e nvim {file} +{line}"
+--
+-- where `{file}` gives a file name, and `{line}` gives the line number. This
```
aidanholm: Replace with: "Before running the command, `{file}` will be replaced with the file to edit, and `line` will be replaced with the line number at which to begin editing."

lib/editor.lua
```diff
@@ -1,16 +1,60 @@
 --- Text editor launching functionality.
 --
 -- This module is primarily for use by other Lua modules that wish to
--- allow the user to edit a particular text file. This module does not
--- have any user-facing modifications or features.
+-- allow the user to edit a particular text file. The default is to guess at the
+-- shell command to open a text editor from environment variables. To override
+-- the guess, replace `editor.cmd_string`. This can be done manually, as follows:
+--
+--     editor = require "editor"
```
aidanholm: Make that `local editor`, and ditto for following code block

aidanholm:
A few more minor nitpicks, but it's looking pretty good!

One other thing is to delete the "Can't yet use a graphical text editor (terminal only)." line, as it's no longer accurate.

gleachkr:
Great, thanks! I'll fix these ASAP. I think it might have been absorbed when the diff became outdated, but I was also having some issues with the code-block syntax. Details at https://github.com/luakit/luakit/pull/361#pullrequestreview-52175615. Should I ignore this for now?

lib/editor.lua
aidanholm: Oops; there's a special condition in the documentation generator to syntax-highlight the code if `lxsh` is installed; looks like I forgot to test it properly...

aidanholm:
Should be fixed at latest `develop`

gleachkr:
It is indeed fixed. I've handled those new comments, rebased and updated the changelog. How's it looking?

lib/editor.lua
```diff
@@ -1,33 +1,76 @@
...
+_M.editor_cmd = _M.builtin.default
+
 --- Edit a file in a terminal editor in a new window.
 --
 -- * Can't yet handle files with special characters in their name.
--- * Can't yet use a graphical text editor (terminal only).
 -- * Can't determine when text editor is closed.
 --
 -- @tparam string file The path of the file to edit.
 -- @tparam number line The line number at which to begin editing.
```
aidanholm: `@tparam` should be `@tparam[opt]`

lib/editor.lua
```diff
@@ -1,33 +1,76 @@
...
+--
+--     local editor = require "editor"
+--     editor.editor_cmd = editor.builtin.urxvt
 --
 -- @module editor
+-- @copyright 2017 Graham Leach-Krouse
+-- @copyright 2017 Aidan Holm
 
-local globals = require("globals")
 local capi = { luakit = luakit }
```
aidanholm: Remove the capi line, and then change `capi.luakit` below to just `luakit`.

aidanholm:
Two more nitpicks, and that should be it; separately, fixup fa4dd99 onto the other commits.

gleachkr:
OK, that should do it. Thanks for your comments and patience.

aidanholm:
Looks great! Thanks for your hard work, and patience with all my nitpicking!

PR #385, opened 2017-07-29
Allow functions in formfiller fields

gleachkr:
Here's an initial PR. I'll add something for extending the DSL in a moment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
  - Added status bar widget that notifies of any Lua warnings or errors.
  - Added migration guide to documentation.
  - Added `luakit.wch_upper` and `luakit.wch_lower` key case conversion utility functions.
+ - Added `formfiller.extend` function for extending the formfiller DSL.
 
 ### Fixed
 
--- a/lib/formfiller.lua
+++ b/lib/formfiller.lua
@@ -135,6 +135,45 @@ local DSL = {
     end,
 }
 
+local dsl_extensions = {}
+
+formfiller_wm:add_signal("dsl_extension_query", function (_, k, arg, view_id)
+    local reply = dsl_extensions[k](unpack(arg))
+    formfiller_wm:emit_signal(view_id,"dsl_extension_reply", reply, view_id)
+end)
+
+--- Extend the formfiller DSL with additional functions. This takes a table of
+-- functions, for example
+--
+--     {pass = function(s) return io.popen("pass " .. s):read()}
+--
+-- which will then be usable in fields of `form.lua`. For example:
+--
+--     input {
+--         name = "username",
+--         value = pass("emailpassword"),
+--     }
+--
+-- functions used to extend the DSL will be called only when needed: when
+-- matching for attributes used in matching, or once a form is applied, for
+-- attributes used in form application.
+--
+-- @tparam table extensions The table of functions extending the formfiller DSL.
+_M.extend = function (extensions)
+    for k, v in pairs(extensions) do
+        assert(type(v) == "function", "bad DSL extension: values must be functions")
+        assert(type(k) == "string", "bad DSL extension: keys must be strings")
+        assert(k ~= "on" and k ~= "form" and k ~= "input", "bad DSL extension: don't shadow core DSL functions")
+    end
+    dsl_extensions = extensions
+    for k, _ in pairs(extensions) do
+        DSL[k] = function (_, ...)
+                return {sentinel = true, arg = {...}, key = k}
+        end
+    end
+    return true
+end
+
 --- Reads the rules from the formfiller DSL file
 local function read_formfiller_rules_from_file()
     local state = {
@@ -256,6 +295,9 @@ webview.add_signal("init", function (view)
         local rules = read_formfiller_rules_from_file()
         local form_specs = form_specs_for_uri(rules, v.uri)
         for _, form_spec in ipairs(form_specs) do
+            if type(form_spec.autofill) == "table" and form_spec.autofill.sentinel then
+                form_spec.autofill = dsl_extensions[form_spec.autofill.key](unpack(form_spec.autofill.arg))
+            end
             if form_spec.autofill then
                 -- Precaution: pattern must contain full domain of page URI
                 local domain = lousy.util.lua_escape(lousy.uri.parse(v.uri).host .. "/")
--- a/lib/formfiller_wm.lua
+++ b/lib/formfiller_wm.lua
@@ -122,30 +122,52 @@ local stylesheet = [===[
 }
 ]===]
 
-local function apply (form, form_spec)
-    -- Map of attr -> value that form has to match
-    local attrs = {}
-    for _, v in ipairs({"method", "name", "id", "action", "className"}) do
-        attrs[v] = form_spec[v]
-    end
-    if not element_attributes_match(form, attrs) then
-        return false
-    end
+local dsl_coroutines = {}
 
-    for _, input_spec in ipairs(form_spec.inputs) do
-        local matches = match("input", {"name", "id", "className", "type"}, input_spec, {form})
-        if #matches > 0 then
-            local val = input_spec.value or input_spec.checked
-            if val then fill_input(matches, val) end
-            if input_spec.focus then matches[1]:focus() end
-            if input_spec.select then matches[1]:select() end
+ui:add_signal("dsl_extension_reply", function(_, _, v, view_id)
+    coroutine.resume(dsl_coroutines[view_id],v)
+end)
+
+local function traverse(view_id, t)
+    if type(t) == "table" and t.sentinel then
+        ui:emit_signal("dsl_extension_query", t.key,t.arg,view_id)
+        return coroutine.yield(dsl_coroutines[view_id])
+    elseif type(t) == "table" then
+        for k,v in pairs(t) do t[k] = traverse(view_id, v) end
+    end
+    return t
+end
+
+local function apply (form, form_spec, page)
+    local co = coroutine.create(function ()
+        -- Map of attr -> value that form has to match
+        local attrs = {}
+        for _, v in ipairs({"method", "name", "id", "action", "className"}) do
+            attrs[v] = traverse(page.id, form_spec[v]) -- traverse and evaluate attributes
+            form_spec[v] = attrs[v] -- write them back to the form_spec
         end
-    end
-    if form_spec.submit then
-        submit_form(form, type(form_spec.submit) == "number" and form_spec.submit or 0)
-    end
+        if not element_attributes_match(form, attrs) then
+            return false
+        end
+        traverse(page.id, form_spec) -- traverse the rest of the form_spec
+        for _, input_spec in ipairs(form_spec.inputs) do
+            local matches = match("input", {"name", "id", "className", "type"}, input_spec, {form})
+            if #matches > 0 then
+                local val = input_spec.value or input_spec.checked
+                if val then fill_input(matches, val) end
+                if input_spec.focus then matches[1]:focus() end
+                if input_spec.select then matches[1]:select() end
+            end
+        end
+        if form_spec.submit then
+            submit_form(form, type(form_spec.submit) == "number" and form_spec.submit or 0)
+        end
 
-    return true
+        dsl_coroutines[page.id] = nil
+        return true
+    end)
+    dsl_coroutines[page.id] = co
+    coroutine.resume(co)
 end
 
 local function formfiller_fill (page, form, form_specs)
@@ -153,7 +175,7 @@ local function formfiller_fill (page, form, form_specs)
     assert(type(form) == "dom_element" and form.tag_name == "FORM")
 
     for _, form_spec in ipairs(form_specs) do
-        if apply(form, form_spec) then
+        if apply(form, form_spec, page) then
             break
         end
     end


gleachkr:
OK, that's a first pass.

lib/formfiller.lua
```diff
@@ -139,6 +143,33 @@ local DSL = {
...
+--         value = pass("emailpassword"),
+--         }
+--
+-- note that functions used to extend the DSL will be called only when
+-- a matching form is applied.
+--
+-- @tparam table extensions The table of functions extending the formfiller DSL
+_M.extend = function (extensions)
+    for k,v in pairs(extensions) do
+        if type(v) == "function" then
```
aidanholm: This should probably be an assertion

lib/formfiller.lua
```diff
@@ -139,6 +143,33 @@ local DSL = {
...
+            DSL[k] = function (_,...)
```
aidanholm: What should happen if there's already something in the DSL table with that key?

lib/formfiller.lua
aidanholm: Should assert `type(k) == "string"` too

lib/formfiller.lua
```diff
@@ -139,6 +143,33 @@ local DSL = {
...
+                local arg = {...}
+                return function () return v(unpack(arg)) end
```
aidanholm: This is good!

lib/formfiller.lua
```diff
@@ -139,6 +143,33 @@ local DSL = {
...
+-- which will then be usable in fields of `form.lua`, for example
+--
+--     input {
+--         name = "username",
+--         value = pass("emailpassword"),
+--         }
+--
+-- note that functions used to extend the DSL will be called only when
+-- a matching form is applied.
+--
```
aidanholm: - Should add a note that the function's environment is discarded, but its closure is kept.
- Should add a note about always executing functions if they're one of the form attributes used to match forms.
- Should add a note that if the function returns another function, that'll be called to get the final value.

lib/formfiller.lua
```diff
@@ -52,7 +52,7 @@
 --   in the example above) takes a Lua pattern!
 --   BEWARE its escaping!
 --
--- * All of the attributes of the <code>form</code> and <code>input</code> tables
+-- * All of the attributes of the <code>form</code> and <code>input</code> tables (
```
aidanholm: Superfluous `(`?

lib/formfiller.lua
```diff
@@ -62,6 +62,14 @@
 --   critically important, however, to verify that the URI pattern of the rule is
 --   correct!
 --
+-- * A field of the form or input tables may also be filled with a function.
+--   When the form is applied, this function will be called without any
+--   arguments, and its result will be used as the entry for the relevant field.
```
aidanholm: - Returned value, not result.
- Should add a note that if the function returns another function, that'll be called to get the value.
  - Not sure if I like that actually; perhaps tweak the traverse function to get rid of that

lib/formfiller_wm.lua
```diff
@@ -122,16 +122,25 @@ local stylesheet = [===[
 }
 ]===]
 
+local function traverse(t)
+    if type(t) == "function" then return traverse(t()) end
```
aidanholm: Re above points, this line would become:
```lua
if type(t) == "function" then return setfenv(t, {})() end
```

lib/formfiller_wm.lua
```diff
@@ -122,16 +122,25 @@ local stylesheet = [===[
...
+    end
+    return t
+end
+
 local function apply (form, form_spec)
     -- Map of attr -> value that form has to match
     local attrs = {}
     for _, v in ipairs({"method", "name", "id", "action", "className"}) do
-        attrs[v] = form_spec[v]
+        attrs[v] = traverse(form_spec[v]) --traverse and evaluate attributes
```
aidanholm: Space after `--`, ditto below

lib/formfiller.lua
gleachkr: Hm. Maybe just don't allow the functions in the canonical DSL to be overwritten? Possibly also throw a warning?

aidanholm: I'd say don't allow overwriting at all; so let's make it return a boolean status code, and check that all keys are available  and do all assertions before doing any assignments to the DSL.

gleachkr:
Just a heads up---driving cross-country today, probably won't be able to finish this up until tomorrow.

aidanholm:
No problem; I'm stuck in a typhoon without power, so I won't be getting much done either

gleachkr:
Thanks for the comments. Updated.

Re: Typhoon. Yikes. Stay safe.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 ### Added
 
  - New `history.frozen` API allows temporarily freezing history collection.
+ - Formfiller forms used in `forms.lua` can now take functions in their fields.
+ - Formfiller DSL is now extensible via `formfiller.extend`.
```
aidanholm: _The_ formfiller DSL...

lib/formfiller.lua
```diff
@@ -135,6 +143,39 @@ local DSL = {
...
+--
+-- note that functions used to extend the DSL will be called only when
+-- needed---when matching for attributes used in matching, or once a form is
+-- applied, for attributes used in form application---and that their
+-- environments are discarded (although their closures are retained).
+--
+-- @tparam table extensions The table of functions extending the formfiller DSL
+_M.extend = function (extensions)
+    for k,v in pairs(extensions) do
+        assert(type(v) == "function", "Bad DSL extension: values must be functions")
```
aidanholm: Don't capitalize "Bad"

lib/formfiller.lua
```diff
@@ -135,6 +143,39 @@ local DSL = {
...
+        assert(type(k) == "string", "Bad DSL extension: keys must be strings")
+        assert(k ~= "on" and k ~= "form" and k ~= "input" , "Bad DSL extension: don't shadow core DSL functions")
```
aidanholm: `"input" , "Bad` <-- has extra space before comma

lib/formfiller_wm.lua
```diff
@@ -122,16 +122,25 @@ local stylesheet = [===[
 }
 ]===]
 
+local function traverse(t)
+    if type(t) == "function" then return setfenv(t,{unpack = unpack})() end
```
aidanholm: Hm; it'd be desirable to avoid this, perhaps by making `unpack` an upvalue of the function, but I have no idea what happens if you serialize built-in C functions... perhaps try and see?

lib/formfiller.lua
```diff
@@ -62,6 +62,14 @@
 --   critically important, however, to verify that the URI pattern of the rule is
 --   correct!
 --
+-- * A field of the form or input tables may also be filled with a function.
+--   When the form is applied, this function will be called without any
+--   arguments, and its returned value will be used as the entry for the relevant
+--   field.  Because of the way that the formfiller DSL is sandboxed, most
```
aidanholm: Be specific: the function is called with an empty environment, and with all upvalues replaced by copies of their values at the function instantiation time.

aidanholm:
I've been thinking, and perhaps there's a better way than serializing the functions. The idea is that when the formfiller needs to evaluate a custom function on on the web process, it pauses formfilling and sends a message to the UI process, which looks up the DSL extension function, calls it, and sends the response back to the web process. The web process then continues filling in the form.

This removes all restrictions on the extension function, which makes the API much easier to use.

gleachkr:
No worries, it's a learning experience.

So, the idea might be to alter `formfiller.extend` in such a way that it saves the extension table and sets `DSL[k]` to a function that calls over to the UI process and asks it to look up `k` in the extension table, evaluate `v` and send back the result?

I might need some pointers for how to get the processes to talk to one another. Where should I look for that?

lib/formfiller_wm.lua
gleachkr: It looks like you can't serialize them. The result is that unpack just returns `nil` on the other side. However, if you write a pure lua unpack, you can use that as an upvalue and keep the fenv clean.

gleachkr:
Ok, I've got a proof of concept. I'll clean it up and push tomorrow.

aidanholm:
That's right, except for the bit about setting `DSL[k]` to a function that does the IPC. I'd prefer some kind of `(sentinel, argument)` pair that indicates an IPC call is needed. Maybe something as simple as `{ argument = s, sentinel = true }`. You'd then check for `type(t) == "table" and t.sentinel` in `traverse()`.

The entire operation of "filling a form" will need to be wrapped in a coroutine that represents the form filling task, as the entire task will need to be paused during the IPC call(s) and resumed after the response arrives.

gleachkr:
OK, reworked to use the the sentinel/argument pattern and to evaluate functions in the UI process.

lib/formfiller_wm.lua
```diff
@@ -122,38 +122,61 @@ local stylesheet = [===[
...
 
-local function apply (form, form_spec)
-    -- Map of attr -> value that form has to match
-    local attrs = {}
-    for _, v in ipairs({"method", "name", "id", "action", "className"}) do
-        attrs[v] = form_spec[v]
-    end
-    if not element_attributes_match(form, attrs) then
-        return false
+local dsl_coroutine
```
aidanholm: This mostly looks good, but a single global coroutine is insufficient. To be sufficiently robust, you will need a table of view_ids -> coroutines, as there may be more than one form-filling operation in progress for a single web process. This doesn't happen very often, of course, but could easily happen during testing or when other modules use the formfiller.

See `lib/chrome_wm.lua` for an example.

lib/formfiller_wm.lua
```diff
@@ -122,38 +122,61 @@ local stylesheet = [===[
...
+
+ui:add_signal("dsl_extension_reply", function(_,_,v)
```
aidanholm: Spaces after commas

gleachkr:
Good point about the coroutine table. Now fixed.

lib/formfiller_wm.lua
```diff
@@ -122,38 +122,62 @@ local stylesheet = [===[
...
-    if not element_attributes_match(form, attrs) then
-        return false
+local dsl_coroutines = {}
+
+ui:add_signal("dsl_extension_reply", function(_, _, v, view_id)
+    coroutine.resume(dsl_coroutines[view_id],v)
+end)
+
+local function traverse(view_id, t)
+    if type(t) == "function" then
```
aidanholm: With the extensions table, I'd prefer that support for functions in the formfiller file be removed. It's simpler to have one way to do things, and functions using the extension table are much less restricted. You'll need to remove the large doc comment, as it refers to this functionality.

lib/formfiller.lua
```diff
@@ -62,6 +62,17 @@
 --   critically important, however, to verify that the URI pattern of the rule is
 --   correct!
 --
+-- * A field of the form or input tables may also be filled with a function.
```
aidanholm: Remove this comment; see below.

lib/formfiller.lua
```diff
@@ -135,6 +146,46 @@ local DSL = {
...
+    local reply = dsl_extensions[k](unpack(arg))
+    formfiller_wm:emit_signal(view_id,"dsl_extension_reply", reply, view_id)
+end)
+
+--- Extend the formfiller DSL with additional functions. This takes a table of
+-- functions, for example
+--
+--     {pass = function(s) return io.popen("pass " .. s):read()}
+--
+-- which will then be usable in fields of `form.lua`, for example
```
aidanholm: should be "`form.lua`. For example:"

lib/formfiller.lua
```diff
@@ -135,6 +146,46 @@ local DSL = {
...
+--
+--     input {
+--         name = "username",
+--         value = pass("emailpassword"),
+--         }
```
aidanholm: Trailing brace should be aligned with 'i' of 'input'.

lib/formfiller.lua
```diff
@@ -135,6 +146,46 @@ local DSL = {
...
+--
+-- functions used to extend the DSL will be called only when needed---when
```
aidanholm: Capital 'f' for functions.

aidanholm: Replace `---` with `: `

lib/formfiller.lua
```diff
@@ -135,6 +146,46 @@ local DSL = {
...
+-- matching for attributes used in matching, or once a form is applied, for
+-- attributes used in form application. DSL extensions are evaluated in the
```
aidanholm: Can remove the line about evaluation in the UI process env; no need to state it if there aren't two mechanisms

lib/formfiller.lua
```diff
@@ -135,6 +146,46 @@ local DSL = {
...
+-- environment of the UI process.
+--
+-- @tparam table extensions The table of functions extending the formfiller DSL
```
aidanholm: Should end in full stop

gleachkr:
No worries. Getting rid of the redundant support for directly inserting functions as field values seems reasonable. Here are the changes. I also made a tweak to allow formfiller extensions in the autofill field, and combined the first two commits, since the first was no longer doing anything.

aidanholm:
And, it's in! Thanks for your hard and patient work!

PR #399, opened 2017-08-02
FreeBSD socket cleanup workaround

xkpd3:
This shall not be a final fix. Up to now I was not able to figure out why the socket on FreeBSD is missing one character. This fix would at least keep /tmp from getting polluted over time.


aidanholm:
Thanks!

Issue #402, opened 2017-08-02
FreeBSD IPC socket path is truncated before creation


aidanholm:
Workaround merged: #399

PR #403, opened 2017-08-02
Fixing the IPC socket truncation

xkpd3:
Right after I posted the workaround, I found the real solution. `offsetof(struct sockaddr_un, sun_path)` should be used to determine the lenth due to different interpretation of null terminated socket paths.

I found the solution here:
https://dev.haiku-os.org/ticket/3255

This should fix https://github.com/luakit/luakit/issues/402


aidanholm:
That was fast, thanks! Closes #402

PR #418, opened 2017-08-09
Fix linker error when running tests

lluixhi:
Source file needs to be before LDFLAGS or `-Wl,--as-needed` will drop symbols
```
 luajit: error loading module 'tests.util' from file './tests/util.so':
        Error relocating ./tests/util.so: g_dir_make_tmp: symbol not found
stack traceback:
        [C]: at 0x9d264f8370
        [C]: in function 'require'
        tests/run_test.lua:15: in main chunk
        [C]: at 0x9d264d12e0
make: *** [Makefile:125: run-tests] Error 1
```


aidanholm:
Thanks for the fix!

Issue #198, opened 2013-04-20
Update jquery


nmeum:
You are still using jquery v1.7.2 (the file is `lib/jquery.min.js`) it is only used for the help, history, bookmark and download page so it shouldn't be too hard to upgrade to jquery v1.9.X or even jquery v2.X, no?

aidanholm:
Hi, and thanks for reporting this!

 Currently, there's no urgent need to update jQuery, as there aren't any (known) bugs or performance issues affecting luakit's chrome pages. Additionally, I'm planning to refactor the chrome pages you mentioned and remove jQuery entirely in the process of doing so. This is because it's the only JS file in `lib/`, it's quite large, and it can be removed without any loss of functionality.

I'll close this now; feel free to reopen if you have questions or anything like that.

hippi777:
hi! :)

i couldn't test all of it out, the toggle works, but without animation, but i have no associated text editor, but can be a good start, i've seen there are other 3 files using jq, but currently i had time only for this much, but i think the rest will be this much easier if they are similar...

as it's not 100% the same as the original in function, test it twice :)

lib/introspector_chrome.lua
```javascript
window.onload=function () {
    // toggle source display
    let bindClass = document.getElementsByClassName("bind");
    for (let b = 0; b < bindClass.length; b++) {
        bindClass[b].onclick = function (event) {
            let func_sourceStyle = bindClass[b].getElementsByClassName("func-source")[0].style;
            if (func_sourceStyle.display == "none") {
                func_sourceStyle.display = "block";
            } else {
                func_sourceStyle.display = "none";
            }
        }
    }

    // open a file
    let filenameClass = document.getElementsByClassName("filename");
    for (let f = 0; f < filenameClass.length; l++) {
        filenameClass[f].onclick = function (event) { //a <span> with "clickable" class would be much better instead of using <a> tags
            event.preventDefault(); //not needed with using <span>
            open_editor(event.target.innerText);
            return false; //not necessary at all?
        }
    }

    // open a file with a custom line
    let linedefinedClass = document.getElementsByClassName("linedefined");
    for (let l = 0; l < linedefinedClass.length; l++) {
        linedefinedClass[l].onclick = function (event) {
            event.preventDefault();
            let t = event.target;
            open_editor(t.getAttribute("filename"), t.getAttribute("line")); //should be data-filename - search&replace can help
            return false;
        }
    }
}
```

i even adopted the style (far from my own...), but i had no much idea what kinda naming convention would be good :D

[...]

PR #438, opened 2017-08-13
[WIP] partial replacement for jquery

hippi777:
needs testing

a &lt;span&gt; tag with "clickable" class (for the appearance, but maybe as well for the js - so iterate them, and where data-filename appears, that can be made functional) (or a &lt;button&gt; tag) would be much better than using &lt;a&gt; tags
preventDefault and return false are not needed this way, but takes a deeper but easy modification...

custom attributes should have "data-" prefix for valid html5 - a search&replace can help for this

origin: https://github.com/luakit/luakit/issues/198#issuecomment-321867319

### DIFF
--- a/lib/introspector_chrome.lua
+++ b/lib/introspector_chrome.lua
@@ -1,11 +1,7 @@
 --- Provides luakit://introspector/ page.
 --
--- This module provides the luakit://introspector/ page. It is useful for
--- viewing all keybindings and modes on a single page, as well as searching for
--- a keybinding for a particular task.
---
 -- @module introspector_chrome
--- @copyright 2016 Aidan Holm <aidanholm@gmail.com>
+-- @copyright 2016 Aidan Holm
 -- @copyright 2012 Mason Larobina <mason.larobina@gmail.com>
 
 local lousy = require("lousy")
@@ -16,7 +12,7 @@ local history = require("history")
 local markdown = require("markdown")
 local editor = require("editor")
 local get_modes = require("modes").get_modes
-local add_cmds = require("modes").add_cmds
+local add_cmds = require("binds").add_cmds
 
 local _M = {}
 
@@ -69,6 +65,7 @@ local html_template = [==[
         pre {
             margin: 0;
             padding: 0;
+            overflow-wrap: break-word;
         }
 
 
@@ -84,6 +81,10 @@ local html_template = [==[
         .mode .binds {
             clear: both;
             display: block;
+        }
+
+        .hidden {
+            display: none;
         }
 
         .bind {
@@ -145,10 +146,11 @@ local html_template = [==[
         }
 
         .bind pre {
-            padding: 1rem 1.2rem;
-            border-left: 2px solid #69c;
-            background: #f5f7f9;
-            color: #334;
+            margin: 1em;
+            padding: 0.5em;
+            background-color: #EFC;
+            border-top: 1px solid #AC9;
+            border-bottom: 1px solid #AC9;
         }
 
         .bind pre code {
@@ -181,7 +183,7 @@ local html_template = [==[
 </head>
 <body>
     <header id="page-header">
-        <h1>Luakit Introspector</h1>
+        <h1>Luakit Help</h1>
     </header>
     <div class="content-margin">
         {sections}
@@ -193,6 +195,7 @@ local html_template = [==[
         {javascript}
     </script>
 </body>
+</html>
 ]==]
 
 local mode_section_template = [==[
@@ -209,8 +212,8 @@ local mode_section_template = [==[
 local mode_bind_template = [==[
     <li class="bind bind_type_{type}">
         <div class="link-box">
-            <a href="#" class="filename">{filename}</a>
-            <a href="#" class="linedefined" filename="{filename}" line="{linedefined}">{linedefined}</a>
+            <a href="#" class="openeditor" data-filename="{filename}">{filename}</a>
+            <a href="#" class="openeditor" data-filename="{filename}" data-line="{linedefined}">{linedefined}</a>
         </div>
         <hr class="clear" />
         <div class="key">{key}</div>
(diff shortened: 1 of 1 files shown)


lib/introspector_chrome.lua
```diff
@@ -223,28 +223,41 @@ local mode_bind_template = [==[
...
-
-    $body.on("click", ".bind", function (e) {
-        var $src = $(this).find(".func-source");
-        if ($src.is(":visible"))
-            $src.slideUp();
-        else
-            $src.slideDown();
-    })
-});
+window.onload=function () {
```
aidanholm: Use `window.add_event_listener("load", ...)` instead

lib/introspector_chrome.lua
```diff
@@ -223,28 +223,41 @@ local mode_bind_template = [==[
...
+    // toggle source display
+    let bindClass = document.getElementsByClassName("bind");
```
aidanholm: Instead of performing separate queries to get all the elements, then iterating through them, add a single `click` event handler to `document.body`. Then check the clicked element `event.target` with `.matches(selector)` for each of the three selectors in the original code.

lib/introspector_chrome.lua
```diff
@@ -223,28 +223,41 @@ local mode_bind_template = [==[
...
+    for (let b = 0; b < bindClass.length; b++) {
+        bindClass[b].onclick = function (event) {
+            let func_sourceStyle = bindClass[b].getElementsByClassName("func-source")[0].style;
+            if (func_sourceStyle.display == "none") {
```
aidanholm: Use `element.classList.toggle('hidden')`; add a `.hidden` class to the stylesheet if it's not already there.

hippi777:
so i don't know if open_editor works without line, but can be defaulted to 1

for iterating class lists i had problems with the for+in syntax previously, because these are virtual array-s or whatever, and they misbehaved, but that's probably task dependent...

the toggle works without animation

i have no associated text editor so i failed to test those, but the best method for that would be to iterate class="clickable" elements and where data-filename appears, those can be changed, where data-linedefined appears, there that can be used, or otherwise 1 for lineno if open_editor needs the 2nd argument (btw i'd prefer data-lineno or line or row or whatever instead of linedefined)

if you want it i can refine the code this way, but i'm not feeling the temptation currently to modify the whole source... but if you don't feel just as well, then later today maybe i can take some time for that :D

aidanholm (changes requested):

aidanholm:
It's worth noting that you don't have to bother making your JS cross-browser-compatible; you can use anything webkit supports

hippi777:
i know :D but for+in is not about that and data- prefix is just for valid html5 that's never wrong :D

aidanholm:
Sorry, didn't see your second comment before.

No need to use a loop at all with the changes I requested. Toggle animation can be done with CSS. Using `data-` prefix is a good idea. You can use `line` instead of `linedefined` if you prefer. `open_editor`'s line number is optional.

hippi777:
so what should i do and what should have done at all (those parts that i may or may not have enough time for...)

aidanholm:
First work on the requested changes; that's the important bit. If you have enough time you can do the other bits.

hippi777:
i didn't get it where the datas comes from so i can't modify the classes...

window.add_event_listener("load", ...) - why this could be better?

.matches(selector) - i've never seen this, maybe doesn't even exists :D but o1===o2 can be good i think

where to add the .hidden? i mean introspector_chrome.lua or help_chrome.lua or the same with .html would be an unrelated generated doc right?

where can i find the contents of the page? i can't see in any file what i've seen on luakit://introspector... i dind't even really understood the codes :D and i've thought it will be kinda as fast as write what i've made and open an issue :D but fine i wanted the pain :) aand how to assign an editor, to become abled to test the stuffs?

i guess i'll make a simple .hidden {display: none} and let the css3 magic and data- prefixes for others or for myself later, but if not today, then much later, because it seems like i will have time and be online around thursday next time and i have to finish my work because testing will start on monday

and i've found a bug, that for example the F block on the introspector have long lines and they disappear in nowhere (i've got no scrollbars, i really didn't have enough time to make it a home sweet home... but a wrap: word-wrap or something like can solve that, but maybe i wrote wrong code, but should be something like this...)

so i don't know how much i'll do today if any, but hopefully this started as not urgent :D

hippi777:
i've made some mess, because as i wanted to test it i just used an older version, but it's enough for today, it don't wanna work, but u can fix it if you won't wait for me... previously the toggle was working, but the editor with the original jquery didn't came, but instead of that the toggle was activated, this implementation should work i think - when it will work... btw i've found everything i was looking for previously :)

aidanholm:
No problem; I'll review it whenever it's ready. Glad you found everything :)

aidanholm:
Note that I won't be able to merge any changes against the old version of `help_chrome`; it has to be against a reasonably recent version.

hippi777:
sure, i know, therefore i said the mistake to don't do :D

but i still have that hacky messed installation and i've got no automated installer that gets the new sauce, compiles and installs for a click, but i wanna do that 1st, and then i can put it from my test dir to my install dir (because i've putted it into Szoszi's dir, as probably i will reuse my functions in luakit when i will be more familiar with it...) however i have to refine a bit my basic install because it only used /usr/local/ as everything goes there from lua, luajit, lfs and tekui...

the current installation is from a previous sys as now everything works out of the box just with a folder that can be anywhere; installation works without privileges; it only needs the same architecture without reinstallation, and the installer needs sudo/root for a few packages (gcc, make, git, mercurial, libreadline6-dev, libncurses5-dev, libfreetype6-dev, libxft-dev, libxext-dev and libxxf86vm-dev) so things that are needed for a fresh raw debian but not after compiling :)

luakit with dev path can be replaced finely anywhere? (currently i modified prefix, therefore it became hacky and runs from the sauce folder with a bash script for cd + env varables :D ) and can luakit leave untouched folders like xdg /etc and whatever used by luakit, or can dev paths make those local what are there? - so what i wanna achieve is to put it in my folder and make it completely fs independent, and this way everything could work without privileges in home folder and can be moved anywhere (so  a poweruser toolset that works anywhere like on an unprivileged host or from a pendrive...) currently i have a supervisor for my app, that making it place independent by passing basedir and at the very beginning i'm rewriting package.path and package.cpath... i hope the same can be achieved in luakit via env variables...

and thx for any help! :)

ghost:
Hi @szbnwer!

I would like to help. :)
Are you still working on this? If so, which chrome pages have you already finished?

aidanholm:
Hey @taiite; as far as I know, @szbnwer has only worked on migrating the introspector page, and hasn't had time to begin work on any of the others. I know the history, downloads, and bookmarks pages all use jQuery at the moment, so any of those are up for grabs. No need to do all of them at once, either; one at a time is great too.

You'd need to open a separate PR of course. Also, see the review comments on this PR about the preferred ways of using standard JS functions like `add_event_listener`, `el.classList.toggle`, etc. Thanks! :)

hippi777:
hi! :) fine thx! so i've messed with an older version in my pr, the toggle works but just hide/display instead of rolling, i couldn't associate a text editor, so that parts is questionable... check out the diff for changes, the js, maybe data- prefixes on the extra attributes and a removed extra html closing tag was all i've made iirc

now it seems that maybe i will have some more freedom but now there are a s#itload of incoming stuffs are waiting for me and some tasks here and there, after i've done with those and talked with my boss (this makes dependent my free time, but probably i've done there...)

aidanholm:
Closing, as jQuery has been removed already

Issue #392, opened 2017-08-01
Add no-binding-matched notification 


aidanholm:

aidanholm:
@gleachkr Actually I had to revert the PR, as it breaks the insert/passthrough mode notification and passthrough mode; e.g. `Ctrl-Z Ctrl-A` no longer works.

gleachkr:
Rats. OK, I'll see if I can fix that.

aidanholm:
Hm.. I think this might require some refactoring/overhauls. The use of `w:notify()` to indicate the current mode is a hack, and if you try to work around that, your patch will get really ugly.

I'm leaning towards adding a status-bar widget on the far left to indicate the current mode. Thoughts?

gleachkr:
Sure, seems reasonable. Might also be nice in principle to be able to have information in the status bar while also displaying a mode indicator.

aidanholm:
My thoughts exactly

aidanholm:
This isn't a high-priority feature by the way, so dont worry about getting it finished ASAP or anything. I'm much more interested in possibly squeezing a settings chrome page in before the next release; but after that I might take a look at this one myself

PR #465, opened 2017-08-24
add a bind-match-failure warning to the window widget's hit method

gleachkr:
Here's a first pass at closing #392

### DIFF
--- a/lib/window.lua
+++ b/lib/window.lua
@@ -262,7 +262,14 @@ _M.methods = {
 
         local caught, newbuf = lousy.bind.hit(w, w.binds, mods, key, opts)
         if w.win then -- Check binding didn't cause window to exit
+            local oldbuf = (w.buffer or "") .. key
             w.buffer = newbuf
+            if not (newbuf or caught or w.ibar.input.focused or w:is_mode("insert"))
+                and type(key) == "string" and key:len() == 1 then
+                w:notify("no binding matched for " .. oldbuf)
+            elseif key ~= "Scroll" then
+                w:set_prompt()
+            end
             w:update_buf()
         end
         return caught
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
  - Added `webview` widget `hardware_acceleration_policy` property.
  - Added `webview` widget `allow_file_access_from_file_urls` and `allow_universal_access_from_file_urls` properties.
  - Added `settings` module and APIs. This replaces the `domain_props` module.
+ - Added warning when no binding is matched for input.
 
 ### Changed
 


lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
 
         local caught, newbuf = lousy.bind.hit(w, w.binds, mods, key, opts)
         if w.win then -- Check binding didn't cause window to exit
+            local oldbuf
+            if w.buffer then oldbuf = w.buffer .. key else oldbuf = key end
```
aidanholm: Should probably be:
```lua
local oldbuf = (w.buffer or "") .. key
```

lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
...
             w.buffer = newbuf
+            if not (newbuf or caught) and key:len() == 1 and w:is_mode("normal") then
```
aidanholm: What happens without the `w:is_mode("normal")` check? My concern is for things like `ex-follow` mode; pressing `;` followed by some random key wouldn't show anything in this case.

lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
...
+                w:notify(oldbuf .. " is not a prefix of an existing binding")
```
aidanholm: This can probably just be "No binding matched..."

aidanholm (changes requested):
That's a lot less code than I was expecting!

lib/window.lua
gleachkr: It tends to spam the user with no-binding notifications in the non-normal modes, like insert or command. Maybe this should be configurable?

lib/window.lua
gleachkr: How about `"No binding matched by " .. oldbuf`, so that you can tell if it's a typo?

lib/window.lua
aidanholm: You could try moving it to the signal handler around line 169 of window.lua

lib/window.lua
aidanholm: Good thinking, but s/by/for/

lib/window.lua
gleachkr: Same problem there, I think, though I might not be seeing the idea you're seeing.

Also, I'm not sure in that context how to tell when a key is part of a prefix for a binding, and when it's not---it looks like you get `match` false when the keypress is part of a prefix, and just hasn't matched yet.

gleachkr:
Went ahead and applied those suggestions, also adding an option to make the modes in which the warning is triggered extensible (with default value `{"normal","ex-follow"}`).

aidanholm:
I actually don't think that's a good idea; ideally, it'd just work in _all_ modes. I'd say keep the whitelist private, which means no API break if this gets improved later.

aidanholm:
Instead of checking the current mode, you could check whether the input bar or the current view is focused; it's hacky, but it should work.

gleachkr:
Good idea. Still need a special case for insert mode, but seems to work otherwise. Added to latest commit.

lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
 
         local caught, newbuf = lousy.bind.hit(w, w.binds, mods, key, opts)
         if w.win then -- Check binding didn't cause window to exit
+            local oldbuf = (w.buffer or "") .. key
             w.buffer = newbuf
+
```
aidanholm: remove whitespace in this line

lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
 
         local caught, newbuf = lousy.bind.hit(w, w.binds, mods, key, opts)
         if w.win then -- Check binding didn't cause window to exit
+            local oldbuf = (w.buffer or "") .. key
             w.buffer = newbuf
+            
+            if not (newbuf or caught or w.ibar.input.focused or w:is_mode("insert")) and key:len() == 1 then
```
aidanholm: key can be a number, for mouse button bindings:
```
[    6.764182] E [core/common/lualib]: Lua error: attempt to index local 'key' (a number value)
                 Traceback:
                 (1) lib/window.lua:268 in function hit
                 (2) lib/webview.lua:82 in function [anonymous]
```

aidanholm (changes requested):

lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
...
+                w:notify("no binding matched for " .. oldbuf)
+            else
+                w:set_prompt()
```
aidanholm: With this, any notification will disappear if the user scrolls with the mousewheel, but clicks are ignored
I think scrolling with the mousewheel shouldn't dismiss the notice, and clicks should

gleachkr: Yes, that seems reasonable. I don't have a scrollwheel right now. What keys does a scrollwheel send? or is there a better way of detecting scrollwheel events?

lib/window.lua
gleachkr: Right. Should there be a warning when no mouse button binding is matched (perhaps unless it's just a vanilla click), or should mouse events be ignored?

lib/window.lua
aidanholm: If you have a touchpad, that works the same way. Scrolling is done with synthesized key events with `key = "Scroll"`: see `lib/webview.lua:88`

lib/window.lua
aidanholm: I think ignoring clicks is fine

gleachkr:
OK, updated.

lib/window.lua
```diff
@@ -262,7 +262,14 @@ _M.methods = {
 
         local caught, newbuf = lousy.bind.hit(w, w.binds, mods, key, opts)
         if w.win then -- Check binding didn't cause window to exit
+            local oldbuf = (w.buffer or "") .. key
             w.buffer = newbuf
+            if not (newbuf or caught or w.ibar.input.focused or w:is_mode("insert"))
+                and type(key) == "string" and key:len() == 1 then
+                w:notify("no binding matched for " .. oldbuf)
+            elseif not (key == "Scroll") then
```
aidanholm: `key ~= "Scroll"`

aidanholm (changes requested):
final code nit :)

gleachkr:
Great! Updated along with changelog.

aidanholm:
Merged :tada: Thanks :)

PR #481, opened 2017-08-29
Remove redundant download handler

ghost:
Shall we remove the redundant download handler in `viewpdf.lua`?
It is already defined in `rc.lua` and I don't really see the point to overwrite this in a module.

### DIFF
--- a/lib/viewpdf.lua
+++ b/lib/viewpdf.lua
@@ -31,13 +31,6 @@ downloads.add_signal("download::status", function(dl)
     end
 end)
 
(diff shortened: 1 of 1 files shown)


aidanholm:
`viewpdf.lua` still needs to work correctly even if the user removes that snippet from their `rc.lua` (not sure why they would though), and an extra signal handler isn't doing any harm, especially as it only applies to PDFs; if the user wants to override how PDFs are downloaded/opened, well, there's no reason to use `viewpdf` at all.

Issue #495, opened 2017-09-01
Compilation errors on OpenBSD


hofheinz:
Compilation currently fails on OpenBSD because
 - OpenBSD uses a very old gcc version (GCC 4.2.1 on OpenBSD 6.1) that doesn't understand `-Wunused-errors`
 - for the same reason, `#pragma GCC diagnostic`s inside functions are not supported (this causes trouble in `widgets/webview.c`)
 - OpenBSD's grep is different from Linux's, and the expression `grep -q "^Lua 5\\.1\\|^LuaJIT" in config.mk does not work as expected (and hence, no Lua executables are found automatically).

This is easily fixed by making `-Wunused` errors and the `#pragma`s in `widget/webview.c` conditional on `uname_s!=OpenBSD`, and by slightly changing the RE in `config.mk`. (PR follows.) Once this is done, I experience no problems compiling and running under OpenBSD.

aidanholm:
@qbit any thoughts on this? Are you experiencing any problems on OpenBSD?

qbit:
@hofheinz are you on release or snapshots?

I think the `LUA_BIN_NAMES` change should go in.

Not super sure about the rest as i386 and amd64, the only two arches that luakit builds on currently (in snapshots), have switched to `clang` for base:

```
qbit@slip[0]:sys[BIG]λ cc -v
OpenBSD clang version 4.0.0 (tags/RELEASE_400/final) (based on LLVM 4.0.0)
Target: amd64-unknown-openbsd6.2
Thread model: posix
InstalledDir: /usr/bin
```

The  `#pragma`s will be ignored by `clang`, but at the same token, users would have to go out of their way to build with the old gcc.

hofheinz:
@qbit I'm on release (6.1), and I did not use your patches from ports when compiling. Ah, I didn't realize OpenBSD is switching to `clang`, yes, this changes things of course. So how about I change the PR to just the `LUA_BIN_NAMES` thing? (Or would this be something better placed in the ports patches?)

aidanholm:
I am happy to merge the `LUA_BIN_NAMES` (adding `luajit51` and the `-E` flag). I'd prefer not to mangle the code just to support old versions of GCC if an alternative compilation method is reasonably easy to obtain.

hofheinz:
Sure, makes a lot of sense. The commit/PR is modified.

[...]

PR #496, opened 2017-09-01
Fix compilation under OpenBSD (#495)

hofheinz:
Some ugly bits: `uname_s` is now defined in the non-platform-specific part (which is, strictly speaking, now platform specific -- maybe these parts should be moved down). Also, it's weird to have only one macro `__OpenBSD__`, but none for Linux, FreeBSD, etc. (should they be added? Or is there a better way to detect the OS in cpp?). Speaking of which, I'm not sure what the effects of commenting out the `#pragma`s in `widget/webview.c` are. (Probably nothing too dramatic, and it certainly doesn't work with them in place.) Please let me know if any of this should be changed.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -77,13 +77,13 @@ endif
 
 LUA_BIN_NAMES += lua-5.1 lua5.1 lua51
 ifneq ($(USE_LUAJIT),0)
-	LUA_BIN_NAMES := luajit $(LUA_BIN_NAMES)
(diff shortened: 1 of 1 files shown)


hofheinz:
Besides, maybe it's not even desirable to introduce ugliness in order to comfort ancient `gcc`s. (Currently, there is an OpenBSD port that simply removes the incompatible parts with a few local patches and additions. Those patches do not know about the `#pragma` and the RE, though.)

aidanholm:
Merged; thanks for the PR!

Issue #497, opened 2017-09-01
Option to always show tab bar


hofheinz:
Feature (?) proposal: most other browsers always show the tab bar (even if only one tab is open). The only way I could come up with to have this in `rc.lua` is to force a `w.tablist.widget.visible = true` upon each `page-removed` and `page-added` signal. (This seems not very elegant, though.)

Perhaps a more elegant way might be to have a `window.always_display_tabbar` setting. (Drawback: this might require adding a parameter like `always_visible` to the `tablist` widget constructor, and hence might mean slightly changing internal APIs.)

Is this a useful feature to have (as an additional setting)? If yes, I could offer a PR to that effect.

aidanholm:
This seems interesting; I don't mind adding a setting for this, although I think  it would be better named `tablist.always_visible`.

Seems simple enough: register a setting, and query its value in `update_tablist_visibility()`; PR welcome and appreciated!

hofheinz:
Great! What would you say is cleaner: have the `tablist` widget register and query that setting (currently `tablist` doesn't `require "settings"` yet), or have someone else (`window`?) pass that setting as a parameter to the `tablist` constructor?

aidanholm:
I'd say `tablist` should do that

hofheinz:
I could not find any `tablist` or full settings documentation, so I didn't add anything there. Also, perhaps it should be noted that `lousy.widget.tablist` must be explicitly loaded/`require`d to have the setting available. (It's not enough to load `lousy`.)

aidanholm:
There are docs at luakit://help/doc/modules/lousy.widget.tablist.html but no settings docs yet. I imagine most people will be setting settings through luakit://settings/ (when that's merged) so I won't require full documentation yet.

PR #498, opened 2017-09-01
Add `tablist.always_visible` setting (#497)

hofheinz:


lib/lousy/widget/tablist.lua
```diff
@@ -211,6 +216,13 @@ function _M.new(notebook, orientation)
     return tlist
 end
 
+settings.register_settings({
+    ["tablist.always_visible"] = {
+        type = "boolean",
+        default = false,
```
aidanholm, line 221: This also needs `domain_specific = false,`

lib/lousy/widget/tablist.lua
```diff
@@ -123,8 +124,12 @@ function _M.new(notebook, orientation)
...
     -- Attach notebook signal handlers
     local function update_tablist_visibility()
-        if tlist.visible and notebook:count() >= 2 then tlist.widget:show() end
-        if not tlist.visible or notebook:count() < 2 then tlist.widget:hide() end
+        if tlist.visible and (notebook:count() >= 2 or settings.tablist.always_visible) then
+            tlist.widget:show()
+        end
+        if not tlist.visible or (notebook:count() < 2 and not settings.tablist.always_visible) then
+            tlist.widget:hide()
+        end
```
aidanholm: I'd prefer that this be changed to set `tlist.widget.visible = true/false` instead; it's much clearer:
```lua
if settings.get_setting("tablist.always_visible") then
    tlist.widget.visible = true
else
    tlist.widget.visible = notebook:count() > 1
end
```

aidanholm: Note that to be really robust, we need to call `update_tablist_visibility` when the `"tablist.always_visible"` setting changes, but I haven't added a `setting-changed` signal yet.

lib/lousy/widget/tablist.lua
```diff
@@ -123,8 +124,12 @@ function _M.new(notebook, orientation)
 
     -- Attach notebook signal handlers
     local function update_tablist_visibility()
-        if tlist.visible and notebook:count() >= 2 then tlist.widget:show() end
-        if not tlist.visible or notebook:count() < 2 then tlist.widget:hide() end
+        if tlist.visible and (notebook:count() >= 2 or settings.tablist.always_visible) then
```
aidanholm: Don't use `settings.tablist.always_visible`; this is an API for the user, not for plugins. It won't return the default value, for example, and it's slower (not that that should matter much).

aidanholm (changes requested):
Thanks for the PR; looks good, but some minor changes are needed

hofheinz:
Thank you for the code review and the constructive comments! I have updated the commit (and taken your version of the `update_tablist_visibility` method, works as expected). I agree with your remark about robustness (what if `tablist.always_visible` gets updated while one tab is open?). I could add a "todo" comment at the `update_tablist_visibility` method, and/or I can try to keep an eye on future commits (to see when the right signal has been added and change the method then).

Please allow one question about your version of `update_tablist_visibility`: you always set the `tlist.visible` property (to true or false), even if it already had that same value. In contrast, the previous code (not mine of course) went out of its way to only set `tlist.visible` if that changed its value (from true to false, or from false to true). Was there a reason for that? (Like: even changing from true to true triggers all kinds of then-unnecessary signals or such.)

aidanholm:
I'll probably add the code to listen for setting changes when that's finished; don't worry about it.

The way the old code worked was mostly just because of the way using method calls encourages a particular code style. It shouldn't have any performance impact, and if it does I'd consider that a bug in the widget implementation (which would be easy enough to fix).

hofheinz:
Thanks for the explanation!

aidanholm:
No worries, questions always welcome :)

aidanholm:
Thanks for the PR!

Issue #460, opened 2017-08-23
Add :set command, for changing settings


aidanholm:
Example usage:
```lua
:set webview.zoom_level 150
:set on .github.com webview.zoom_level 130  -- Domain specific setting
:set window.new_window_size 1200x900 -- No need to wrap strings in "quotes"
```

This would be made _much_ better with #461, which would allow saving changes over restarts.

norcalli:
Should this be closed?

Issue #459, opened 2017-08-23
Add luakit://settings/ chrome page


aidanholm:
Basically, a chrome page for viewing and modifying settings, similar to qutebrowser's `qute://settings`.

This would be made _much_ better with #461, which would allow saving changes over restarts.

aidanholm:
Added at a8343c04

PR #502, opened 2017-09-04
Add :set command, for changing settings (#460)

ghost:
Hi!

I added two public methods to settings.lua (set_setting and set_setting_on) to change settings with and without specifying a domain.
Then I added two commands that call these methods: set and seton respectively.

So it works like this:
`:set window.search_engines.default duckduckgo.com/?q=%s` changes the default search engine.
`:seton github.com webview.zoom_level 80` changes the zoom level on github only.

Webview settings need the tab to be refreshed before being applied, I don't know if it's ok.

I've run the tests this time c: but I haven't changed the changelog.

### DIFF
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -595,7 +595,7 @@ modes.add_cmds({
             value = convert(value, setting.type)
             settings.set_setting(key, value, { domain = domain })
         end,
-        format = "{uri} {setting}",
+        format = "{domain} {setting}",
     }}
 })
 
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -544,6 +544,45 @@ end
         end },
 })
 
+local function convert (str, new_type)
+    local convertion_table = {
+        number = tonumber,
+        boolean = function (val)
+            if val == "true" then return true
+            elseif val == "false" then return false
+            else error("'"..val.."' is not a boolean")
+            end
+        end,
+        string = function (val)
+            return val
+        end,
+        enum = function (val)
+            return val
+        end,
+    }
+
+    return convertion_table[new_type](str)
+end
+
+modes.add_cmds({
+    { ":set", "Change a setting.", {
+        func = function (_, o)
+            local setting, value = unpack(split(o.arg))
+            value = convert(value, settings.get_settings_map()[setting].type)
+            settings.set_setting(setting, value)
+        end,
+        format = "{setting}",
+    }},
+    { ":seton", "Change a setting for a specific domain.", {
+        func = function (_, o)
+            local domain, setting, value = unpack(split(o.arg))
+            value = convert(value, settings.get_settings_map()[setting].type)
+            settings.set_setting(setting, value, { domain = domain })
+        end,
+        format = "{uri} {setting}",
+    }}
+})
+
 return _M
 
 -- vim: et:sw=4:ts=8:sts=4:tw=80
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -563,7 +563,10 @@ end
 modes.add_cmds({
     { ":set", "Change a setting.", {
         func = function (_, o)
-            local setting, value = o.arg:match("^%s+(%S+)%s+(.*)$")
+            o.arg = o.arg or ""
+            local setting, value = o.arg:match("^%s*(%S+)%s+(.*)$")
+            assert(setting and value, "Usage: ':set <setting> <value>'")
+            assert(settings.get_settings()[setting], "Setting not found: "..setting)
             value = convert(value, settings.get_settings()[setting].type)
             settings.set_setting(setting, value)
         end,
@@ -571,7 +574,10 @@ modes.add_cmds({
     }},
     { ":seton", "Change a setting for a specific domain.", {
         func = function (_, o)
-            local domain, setting, value = o.arg:match("^%s+(%S+)%s+(%S+)%s+(.*)$")
+            o.arg = o.arg or ""
+            local domain, setting, value = o.arg:match("^%s*(%S+)%s+(%S+)%s+(.*)$")
+            assert(domain and setting and value, "Usage: ':seton <domain> <setting> <value>'")
+            assert(settings.get_settings()[setting], "Setting not found: "..setting)
             value = convert(value, settings.get_settings()[setting].type)
             settings.set_setting(setting, value, { domain = domain })
         end,
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -553,12 +553,8 @@ local function convert (str, new_type)
             else error("'"..val.."' is not a boolean")
             end
         end,
-        string = function (val)
-            return val
-        end,
-        enum = function (val)
-            return val
-        end,
+        string = function (val) return val end,
+        enum = function (val) return val end,
     }
 
     return convertion_table[new_type](str)
@@ -567,16 +563,16 @@ end
 modes.add_cmds({
     { ":set", "Change a setting.", {
         func = function (_, o)
-            local setting, value = unpack(split(o.arg))
-            value = convert(value, settings.get_settings_map()[setting].type)
+            local setting, value = o.arg:match("^%s+(%S+)%s+(.*)$")
+            value = convert(value, settings.get_settings()[setting].type)
             settings.set_setting(setting, value)
         end,
         format = "{setting}",
     }},
     { ":seton", "Change a setting for a specific domain.", {
         func = function (_, o)
-            local domain, setting, value = unpack(split(o.arg))
-            value = convert(value, settings.get_settings_map()[setting].type)
+            local domain, setting, value = o.arg:match("^%s+(%S+)%s+(%S+)%s+(.*)$")
+            value = convert(value, settings.get_settings()[setting].type)
             settings.set_setting(setting, value, { domain = domain })
         end,
         format = "{uri} {setting}",
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -562,24 +562,38 @@ end
 
 modes.add_cmds({
     { ":set", "Change a setting.", {
-        func = function (_, o)
+        func = function (w, o)
             o.arg = o.arg or ""
-            local setting, value = o.arg:match("^%s*(%S+)%s+(.*)$")
-            assert(setting and value, "Usage: ':set <setting> <value>'")
-            assert(settings.get_settings()[setting], "Setting not found: "..setting)
-            value = convert(value, settings.get_settings()[setting].type)
-            settings.set_setting(setting, value)
+            local key, value = o.arg:match("^%s*(%S+)%s+(.*)$")
+            if (key and value) == nil then
+                w:error("Usage: ':set <setting> <value>'")
+                return
+            end
+            local setting = settings.get_settings()[key]
+            if setting == nil then
+                w:error("Setting not found: "..key)
+                return
+            end
+            value = convert(value, setting.type)
+            settings.set_setting(key, value)
         end,
         format = "{setting}",
     }},
     { ":seton", "Change a setting for a specific domain.", {
-        func = function (_, o)
+        func = function (w, o)
             o.arg = o.arg or ""
-            local domain, setting, value = o.arg:match("^%s*(%S+)%s+(%S+)%s+(.*)$")
-            assert(domain and setting and value, "Usage: ':seton <domain> <setting> <value>'")
-            assert(settings.get_settings()[setting], "Setting not found: "..setting)
-            value = convert(value, settings.get_settings()[setting].type)
-            settings.set_setting(setting, value, { domain = domain })
+            local domain, key, value = o.arg:match("^%s*(%S+)%s+(%S+)%s+(.*)$")
+            if (domain and key and value) == nil then
+                w:error("Usage: ':seton <domain> <setting> <value>'")
+                return
+            end
+            local setting = settings.get_settings()[key]
+            if setting == nil then
+                w:error("Setting not found: "..key)
+                return
+            end
+            value = convert(value, setting.type)
+            settings.set_setting(key, value, { domain = domain })
         end,
         format = "{uri} {setting}",
     }}
--- a/lib/completion.lua
+++ b/lib/completion.lua
@@ -12,6 +12,7 @@ local lousy = require("lousy")
 local history = require("history")
 local bookmarks = require("bookmarks")
 local modes = require("modes")
+local settings = require("settings")
 local new_mode, get_mode = modes.new_mode, modes.get_mode
 local add_binds = modes.add_binds
 local escape = lousy.util.escape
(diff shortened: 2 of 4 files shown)


aidanholm:
Hi!

Thanks for the PR! Unfortuately I've got some work on a `set_setting()` function and persistent settings storage that I haven't cleaned up and pushed yet. Most of the PR look like it could be reused however, so I'll give this one a code review when I've cleaned that up.

The `:set` command syntax looks good. I'd probably prefer that they both use `:set` and domain names are detected automatically, but I'm not sure if that's easy to do. Either way, they'll need completion support for settings names/values, which will depend on an API I haven't finished and pushed yet.

ghost:
Ok, no problem.

Did you also start working on the setting page (#459)? I could work on it with my version of `set_setting`.

aidanholm:
Yep, I've started on that one.

The difference between our implementations is that I've added a persistence mechanism so that settings set with `:set` or the chrome page will be kept even over restarts, while your current implementation requires re-setting after a restart. In contrast, the `settings.foo.bar = baz` method isn't persisted, because this is used in `rc.lua`

Slightly off topic: I'm not necessarily convinced that having three different ways to set settings is ideal, but I also think that it's best to accommodate different user groups; it's all still a work in progress asymtotically approaching the ideal settings system :)

aidanholm:
@taiite I've pushed my changes; you'll need to rebase onto the latest `develop`, and then I'll give your code a review.

lib/binds.lua
```diff
@@ -544,6 +544,23 @@ end
...
+        func = function (_, o)
+            local setting, value = unpack(split(o.arg))
+            settings.set_setting(setting, value)
+        end,
+        format = "{setting} {value}",
+    }},
+    { ":seton", "Change a browser setting for a specific domain.", {
+        func = function (_, o)
+            local domain, setting, value = unpack(split(o.arg))
+            settings.set_setting_on(domain, setting, value)
```
aidanholm: This uses `settings.set_setting` as `:set` does, but with an extra third argument, which should be a table with a single key `domain`. (This makes the API more extensible, if we have e.g. pattern matching URIs in the future).

lib/settings.lua
```diff
@@ -187,6 +187,72 @@ _M.get_setting = function (key)
     return S_get(nil, key)
 end
 
+local function convert (str, new_type)
```
aidanholm: Most of these additions aren't needed anymore (sorry)
I've added a `set_setting` API, and conversion should take place in the command function in `lib/binds.lua`; the `set_setting` API expects a correct value type.

lib/settings.lua
```diff
@@ -187,6 +187,72 @@ _M.get_setting = function (key)
...
+        end,
+        ["string"] = function (val)
+            return val
+        end,
+    }
+
+    return convertion_table[new_type](str)
+end
+
+local function search_for_table (key)
```
aidanholm: Not sure if this function is needed; haven't had a close look at it yet...

lib/binds.lua
```diff
@@ -544,6 +544,23 @@ end
         end },
 })
 
+modes.add_cmds({
```
aidanholm: I'd prefer that this be added in `lib/settings.lua` instead. `lib/binds.lua` and `lib/modes.lua` are already a bit of a disaster area with lots of misc. stuff :/

aidanholm:
Actually, posted a few comments anyway. Code quality looks pretty great in general!

To implement completion for settings. you'll need to expose API in `completion.lua`. I see you've already added a placeholder `{setting}` completion format bit; that may not be sufficient:

Completers cannot access the prior parts of the format string, so your `{value}` completer won't be able to limit suggestions to valid values for the setting, and will have to show all possible setting values, even if they don't make sense for that key.

The way around this is to use a `{settings-key-value}` completer to generate completions, like how the `{command}` completer works. It produces cmpletion suggestions where the key is followed by a completion format string specialized for the type of the setting value. For `{command}`, this is based on the `format` field; for `{settings-key-value}` it'll be based on the type of the setting.

Example completion suggestions:
```
webview.hardware_acceleration_policy always
webview.hardware_acceleration_policy never
webview.hardware_acceleration_policy ondemand
tablist.always_visible {setting-value-boolean}
window.home_page {uri}
```
The completion engine will only suggest a word at a time, so the hardware acceleration one will work like you'd expect.

ghost:
Hi! I've read your comments.

I don't mind if I write more code than needed. I'm not very familiar with lua yet (especially with "metatable" things) so I'm trying things out and checking the docs when I don't understand something. :)

So I've rebased `fix-460` on `develop` and dropped my code in `settings.lua` and I will work on the completiong of `:set` and `:seton`, if you don't mind having two commands (maybe we could do another PR to merge them, because I haven't looked at `completion.lua` yet so I don't know how to deal with optional parameters).

I first tried to call `modes.add_cmds` in `settings.lua`. However it would create a require loop: `modes.lua` requires `window.lua` which requires `settings.lua` (so `settings.lua` can't require `modes.lua`, I think?).

ghost:
By the way, it would be great if local functions (those not exported in a module) would be commented. Just one or two short lines explaining what the function does, even without going in the details, would greatly save time figuring out what has been done and whether there should be another function.

What do you think about this? Should I open an issue to talk about it?

ghost:
I've added a completer for the setting key and I'm still figuring out how to make the completion work with the setting values as well.

However the `:seton` command has only completion for the domain. I don't know what could be the problem but I haven't looked into it this much yet.

aidanholm:
I've changed the existing `get_settings()` function to return a map, like your `get_settings_map()`, as I'd rather not have two functions that do basically the same thing.

lib/completion.lua
```diff
@@ -335,6 +336,23 @@ completers.uri = {
     func = function () return { { format = "{history}" }, { format = "{bookmarks}" }, } end,
 }
 
+completers.setting = {
+    header = { "Setting", "Value" },
+    func = function ()
+        local setting_list = settings.get_settings()
+        local ret = {}
+
+        for _, setting in ipairs(setting_list) do
```
aidanholm: This can just be `for k, setting in pairs(settings.get_settings()) do`

lib/completion.lua
```diff
@@ -335,6 +336,23 @@ completers.uri = {
     func = function () return { { format = "{history}" }, { format = "{bookmarks}" }, } end,
 }
 
+completers.setting = {
+    header = { "Setting", "Value" },
+    func = function ()
+        local setting_list = settings.get_settings()
+        local ret = {}
+
```
aidanholm: Remove blank line

lib/completion.lua
```diff
@@ -335,6 +336,23 @@ completers.uri = {
...
+        for _, setting in ipairs(setting_list) do
+            table.insert(ret, {
+                setting.key, tostring(setting.value),
+                format = setting.key,
+            })
+        end
+
```
aidanholm: Remove blank line

lib/settings.lua
```diff
@@ -251,6 +251,19 @@ _M.get_settings = function ()
     return ret
 end
 
+_M.get_settings_map = function ()
```
aidanholm: No longer needed

lib/binds.lua
```diff
@@ -544,6 +544,45 @@ end
...
+local function convert (str, new_type)
+    local convertion_table = {
+        number = tonumber,
+        boolean = function (val)
+            if val == "true" then return true
+            elseif val == "false" then return false
+            else error("'"..val.."' is not a boolean")
+            end
+        end,
+        string = function (val)
```
aidanholm: Put the function definition all on one line: `function (val) return val end,`

lib/binds.lua
```diff
@@ -544,6 +544,45 @@ end
...
+            return val
+        end,
+        enum = function (val)
```
aidanholm: ditto

lib/binds.lua
```diff
@@ -544,6 +544,45 @@ end
...
+        end,
+    }
+
+    return convertion_table[new_type](str)
+end
+
+modes.add_cmds({
+    { ":set", "Change a setting.", {
+        func = function (_, o)
+            local setting, value = unpack(split(o.arg))
```
aidanholm: This won't work when setting strings that have spaces in them; instead do something like this:
```lua
local setting, value = o.arg:match("^%s+(%S+)%s+(.*)$")
```

lib/binds.lua
```diff
@@ -544,6 +544,45 @@ end
...
+            value = convert(value, settings.get_settings_map()[setting].type)
+            settings.set_setting(setting, value)
+        end,
+        format = "{setting}",
+    }},
+    { ":seton", "Change a setting for a specific domain.", {
+        func = function (_, o)
+            local domain, setting, value = unpack(split(o.arg))
```
aidanholm: Ditto

aidanholm (changes requested):
Mostly small code style issues

ghost:
Fixed the code style and changed the `settings.get_settings_map` calls.

lib/binds.lua
```diff
@@ -544,6 +544,41 @@ end
...
+        enum = function (val) return val end,
+    }
+
+    return convertion_table[new_type](str)
+end
+
+modes.add_cmds({
+    { ":set", "Change a setting.", {
+        func = function (_, o)
+            local setting, value = o.arg:match("^%s+(%S+)%s+(.*)$")
```
aidanholm: I'd prefer that this have some error checking with an informative error message. Try running just `:qmark` to see what you get without error checking... ;)

lib/binds.lua
```diff
@@ -544,6 +544,41 @@ end
...
+            value = convert(value, settings.get_settings()[setting].type)
+            settings.set_setting(setting, value)
+        end,
+        format = "{setting}",
+    }},
+    { ":seton", "Change a setting for a specific domain.", {
+        func = function (_, o)
+            local domain, setting, value = o.arg:match("^%s+(%S+)%s+(%S+)%s+(.*)$")
```
aidanholm: Ditto, pattern fix needed here as well

lib/binds.lua
aidanholm: Sorry, I got the pattern wrong. Should be `"^%s*(%S+)%s+(.*)$"`

lib/completion.lua
```diff
@@ -335,6 +336,20 @@ completers.uri = {
     func = function () return { { format = "{history}" }, { format = "{bookmarks}" }, } end,
 }
 
+completers.setting = {
+    header = { "Setting", "Value" },
+    func = function ()
+        local ret = {}
+        for key, setting in pairs(settings.get_settings()) do
+            table.insert(ret, {
+                key, tostring(setting.value),
```
aidanholm, line 345: This wasn't what I had in mind, but it actually works quite well!

lib/completion.lua
```diff
@@ -335,6 +336,20 @@ completers.uri = {
     func = function () return { { format = "{history}" }, { format = "{bookmarks}" }, } end,
 }
 
+completers.setting = {
+    header = { "Setting", "Value" },
```
aidanholm: I'd prefer this said "Current Value"

aidanholm (changes requested):
Sorry for the slow response; looks pretty good, only minor changes needed

ghost:
Hi! I'll also add a change log entry. :)

lib/binds.lua
```diff
@@ -544,6 +544,47 @@ end
...
+    { ":seton", "Change a setting for a specific domain.", {
+        func = function (_, o)
+            o.arg = o.arg or ""
+            local domain, setting, value = o.arg:match("^%s*(%S+)%s+(%S+)%s+(.*)$")
+            assert(domain and setting and value, "Usage: ':seton <domain> <setting> <value>'")
+            assert(settings.get_settings()[setting], "Setting not found: "..setting)
+            value = convert(value, settings.get_settings()[setting].type)
+            settings.set_setting(setting, value, { domain = domain })
+        end,
+        format = "{uri} {setting}",
```
aidanholm: I tried this, and it was a bad idea on my part. Instead, I'd say make this a `{domain}` completer, and then make the completer match `%S+`.

lib/binds.lua
```diff
@@ -544,6 +544,47 @@ end
...
+
+    return convertion_table[new_type](str)
+end
+
+modes.add_cmds({
+    { ":set", "Change a setting.", {
+        func = function (_, o)
+            o.arg = o.arg or ""
+            local setting, value = o.arg:match("^%s*(%S+)%s+(.*)$")
+            assert(setting and value, "Usage: ':set <setting> <value>'")
```
aidanholm: I'd prefer these assertions use `w:error`; `w` is the first argument to the `func` function.

lib/binds.lua
```diff
@@ -544,6 +544,47 @@ end
...
+            assert(settings.get_settings()[setting], "Setting not found: "..setting)
```
aidanholm: I'd prefer `settings.get_settings` is called once and its result assigned to a local variable

aidanholm (changes requested):

lib/binds.lua
taiite: Done.

ghost:
Hi!

I changed the code according to your two first requests.
However, I don't know how to solve the third... :(

I've tried to create a completer like this, but the completion wouldn't work for the settings:

```lua
completers.domain = {
  func = function () return { format = "%S+" } end,
}
```

The settings would complete only with this completer:

```lua
completers.domain = {
  header = { "Domain", "" },
  func = function () return { { "str1", "str2", format = "%S+" } } end,
}
```

But then the domain can only literally be `%S+`.

aidanholm:
Completion func should be:
```lua
function () return { { format = { lit = "%S+", pattern = true } } } end,
```

aidanholm:
Hi! The completion format should be `{ lit = "%S+", pattern = true }`; the completion function should return a single item for the first part of the buffer matching `%S+`

ghost:
Hey, I can't seem to make it work...
I tried this but it doesn't treat `%S+` as a pattern:

``` lua
completers.domain = {
  header = { "Domain" , "" },
  func = function () return {{ "Domain", "", format = {{ lit = "%S+", pattern = true }} }} end,
}
```

I also tried this, but there is no completion at all:

```lua
completers.domain = {
  func = function () return { format = { lit = "%s+", pattern = true } } end,
}
```

aidanholm:
Try this:
```lua

completers.domain = {
    header = { "Domain", "" },
    func = function (buf)
        local prefix = buf:match("^%S+")
        return prefix and {{ prefix, "", format = {{ lit = prefix }} }} or {}
    end,
}
```

ghost:
I understand now! I haven't thought about doing it like that. Thanks for the help.

The tests should run correctly. Do I need to rebase everything into a single commit?

aidanholm:
Oops, sorry, I completely forgot to review this one :/ I've rebased and merged manually into `develop`; thanks for your work!

PR #505, opened 2017-09-06
Change :introspector to :binds

### LINKED ISSUE #504: `:introspector` is broken
Looks like #483 forgot to add a `:binds` command to open `luakit://binds` in a new tab. The old `:introspector` is still there and tries to open a chrome that no longer exists.

jonas-schievink:
Fixes #504


aidanholm:
Thanks for the patches!

Issue #455, opened 2017-08-21
Remove jQuery, and replace with vanilla JS


aidanholm:
Only chrome pages use JS.

aidanholm:
Done at 4672f3eb

PR #515, opened 2017-09-14
[WIP] Remove jQuery, and replace with vanilla JS (#455)

ghost:
Fixing issue #455 :)

Progress:

- [x] binds_chrome.lua
- [X] bookmarks_chrome.lua
- [ ] downloads_chrome.lua
- [X] history_chrome.lua

I've tried some things out (like creating elements manually). I don't know if it's much faster than before.
I also didn't reproduce animations in vanilla javascript as it doesn't bother me this way. You tell me if you want them back, maybe a bit of css would do the trick.

Code style follows [standardjs][1]. I find it clean, and their parsing tools make it easy to fix code-style-related issues.

[1]: https://standardjs.com/

### DIFF
--- a/lib/binds_chrome.lua
+++ b/lib/binds_chrome.lua
@@ -114,10 +114,6 @@ local html_template = [==[
             text-decoration: underline;
         }
 
-        .bind .func-source {
-            display: none;
-        }
-
         .bind .key {
             font-family: monospace, sans-serif;
             float: left;
@@ -177,6 +173,10 @@ local html_template = [==[
         #templates {
             display: none;
         }
+
+        .hidden {
+            display: none;
+        }
     </style>
 </head>
 <body>
@@ -211,7 +211,7 @@ local mode_bind_template = [==[
         <hr class="clear" />
         <div class="key">{key}</div>
         <div class="box desc">{desc}</div>
-        <div class="box func-source">
+        <div class="box func-source hidden">
             <h4>Function source:</h4>
             <pre><code>{func}</code></pre>
         </div>
@@ -234,8 +234,7 @@ window.addEventListener('load', () => {
 
     listen(document.getElementsByClassName('bind'), 'click', event => {
         let src = event.currentTarget.getElementsByClassName('func-source')[0]
-        if (!src) return
-        src.style.display = src.style.display !== 'block' ? 'block' : 'none'
+        if (src) src.classList.toggle('hidden')
     })
 })
 ]=]
--- a/lib/binds_chrome.lua
+++ b/lib/binds_chrome.lua
@@ -221,27 +221,27 @@ local mode_bind_template = [==[
 
 local main_js = [=[
 function listen (elements, eventType, action) {
-  for (let $el of elements) {
-    $el.addEventListener(eventType, action)
-  }
+    for (let $el of elements) {
+        $el.addEventListener(eventType, action)
+    }
 }
 
 window.addEventListener('load', () => {
-  listen(document.getElementsByClassName('linedefined'), 'click', event => {
-    event.preventDefault()
-    let { filename, line } = event.target.dataset
-    open_editor(filename, line)
-  })
+    listen(document.getElementsByClassName('linedefined'), 'click', event => {
+        event.preventDefault()
+        let { filename, line } = event.target.dataset
+        open_editor(filename, line)
+    })
 
-  listen(document.querySelectorAll('.desc a'), 'click', event => {
-    event.stopPropagation()
-  })
+    listen(document.querySelectorAll('.desc a'), 'click', event => {
+        event.stopPropagation()
+    })
 
-  listen(document.getElementsByClassName('bind'), 'click', event => {
-    let src = event.currentTarget.getElementsByClassName('func-source')[0]
-    if (!src) return
-    src.style.display = src.style.display !== 'block' ? 'block' : 'none'
-  })
+    listen(document.getElementsByClassName('bind'), 'click', event => {
+        let src = event.currentTarget.getElementsByClassName('func-source')[0]
+        if (!src) return
+        src.style.display = src.style.display !== 'block' ? 'block' : 'none'
+    })
 })
 ]=]
 
--- a/lib/binds_chrome.lua
+++ b/lib/binds_chrome.lua
@@ -219,23 +219,17 @@ local mode_bind_template = [==[
 ]==]
 
 local main_js = [=[
-function listen (elements, eventType, action) {
-    for (let $el of elements) {
-        $el.addEventListener(eventType, action)
-    }
-}
-
-window.addEventListener('load', () => {
-    listen(document.getElementsByClassName('linedefined'), 'click', event => {
+document.addEventListener('click', event => {
+    if (event.target.matches('.linedefined')) {
         event.preventDefault()
         let { filename, line } = event.target.dataset
         open_editor(filename, line)
-    })
-
-    listen(document.getElementsByClassName('bind'), 'click', event => {
-        let src = event.currentTarget.getElementsByClassName('func-source')[0]
+    } else if (event.target.matches('.bind, .bind *')) {
+        let $el = event.target
+        while ($el && !$el.classList.contains('bind')) $el = $el.parentNode
+        let src = $el.getElementsByClassName('func-source')[0]
         if (src) src.classList.toggle('hidden')
-    })
+    }
 })
 ]=]
 
--- a/lib/binds_chrome.lua
+++ b/lib/binds_chrome.lua
@@ -232,10 +232,6 @@ window.addEventListener('load', () => {
         open_editor(filename, line)
     })
 
-    listen(document.querySelectorAll('.desc a'), 'click', event => {
-        event.stopPropagation()
-    })
-
     listen(document.getElementsByClassName('bind'), 'click', event => {
         let src = event.currentTarget.getElementsByClassName('func-source')[0]
         if (!src) return
--- a/lib/binds_chrome.lua
+++ b/lib/binds_chrome.lua
@@ -206,8 +206,7 @@ local mode_section_template = [==[
 local mode_bind_template = [==[
     <li class="bind bind_type_{type}">
         <div class="link-box">
-            <a href="#" class="filename">{filename}</a>
-            <a href="#" class="linedefined" data-filename="{filename}" data-line="{linedefined}">{linedefined}</a>
+            <a href="#" class="linedefined" data-filename="{filename}" data-line="{linedefined}">{filename}:{linedefined}</a>
         </div>
         <hr class="clear" />
         <div class="key">{key}</div>
--- a/lib/binds_chrome.lua
+++ b/lib/binds_chrome.lua
@@ -187,9 +187,6 @@ local html_template = [==[
         {sections}
     </div>
     <script>
-        {jquery}
-    </script>
-    <script>
         {javascript}
     </script>
 </body>
@@ -210,7 +207,7 @@ local mode_bind_template = [==[
     <li class="bind bind_type_{type}">
         <div class="link-box">
             <a href="#" class="filename">{filename}</a>
-            <a href="#" class="linedefined" filename="{filename}" line="{linedefined}">{linedefined}</a>
+            <a href="#" class="linedefined" data-filename="{filename}" data-line="{linedefined}">{linedefined}</a>
         </div>
         <hr class="clear" />
         <div class="key">{key}</div>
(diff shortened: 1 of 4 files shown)


lib/binds_chrome.lua
```diff
@@ -223,28 +220,29 @@ local mode_bind_template = [==[
...
+}
+
+window.addEventListener('load', () => {
+  listen(document.getElementsByClassName('linedefined'), 'click', event => {
+    event.preventDefault()
+    let { filename, line } = event.target.dataset
+    open_editor(filename, line)
+  })
+
+  listen(document.querySelectorAll('.desc a'), 'click', event => {
```
aidanholm: Is it necessary to block clicks on description links? This behavior works fine in the current implementation.

lib/binds_chrome.lua
```diff
@@ -223,28 +220,29 @@ local mode_bind_template = [==[
...
+    event.stopPropagation()
+  })
+
+  listen(document.getElementsByClassName('bind'), 'click', event => {
+    let src = event.currentTarget.getElementsByClassName('func-source')[0]
+    if (!src) return
```
aidanholm: This should be `if (src) scr.classList.toggle('hidden')`

lib/binds_chrome.lua
```diff
@@ -223,28 +220,29 @@ local mode_bind_template = [==[
...
-
-    $body.on("click", ".bind", function (e) {
-        var $src = $(this).find(".func-source");
-        if ($src.is(":visible"))
-            $src.slideUp();
-        else
-            $src.slideDown();
-    })
-});
+function listen (elements, eventType, action) {
```
aidanholm: I think instead of using a wrapper function to add multiple event listeners, it'd be better to add one event listener to the `<body>` element, and use `event.target.matches()` to check whether it's a line number, etc.

lib/bookmarks_chrome.lua
```diff
@@ -254,240 +254,247 @@ local html_template = [==[
...
-    function make_bookmark(b) {
-        var $b = $(bookmark_html);
-
-        $b.attr("bookmark_id", b.id);
-        $b.find(".title a").attr("href", b.uri).text(b.title || b.uri);
-        $b.find(".date").text(b.date);
-
-        if (b.title)
-            $b.find(".uri").text(b.uri).show();
+function createElementsFromHtml (html) {
```
aidanholm: Should be `createElementsFromHTML`

lib/bookmarks_chrome.lua
```diff
@@ -254,240 +254,247 @@ local html_template = [==[
...
+        updateNavButtons()
+        return
+      }
+
+      results.map(makeBookmark).forEach($b => $results.appendChild($b))
+    })
+  }
+
+  function updateNavButtons () {
+    $next.style.display = resultsLen === limit ? 'inline' : 'none'
```
aidanholm: I'd prefer using `$next.classList.toggle('hidden', condition)` here; same with `$prev`.

lib/bookmarks_chrome.lua
```diff
@@ -254,240 +254,247 @@ local html_template = [==[
...
+    $prev.style.display = parseInt($page.value, 10) > 1 ? 'inline' : 'none'
+  }
+
+  function findBookmarkParent ($el) {
+    let $b = $el.parentNode
```
aidanholm: Minor nit:
```js
function findBookmarkParent ($el) {
  do { $el = $el.parentNode } while ($el && !$el.classList.contains('bookmark'))
  return $el
}
```

lib/bookmarks_chrome.lua
```diff
@@ -254,240 +254,247 @@ local html_template = [==[
...
+      : undefined
+    let title = $editDialog.title.value
+    let uri = $editDialog.uri.value
+    let tags = $editDialog.tags.value
+    let desc = $editDialog.desc.value
+
+    try {
+      bookmarks_add(uri, { title, tags, desc, created })
+    } catch (err) {
+      window.alert(err)
```
aidanholm: What sort of errors would occur here?

lib/downloads_chrome.lua
```diff
@@ -84,187 +84,136 @@ _M.stylesheet = [==[
...
-function update_list_finish(downloads) {
-    console.log(downloads)
-    // return if no downloads to display
-    if (downloads.length === "undefined") {
-        setTimeout(update, 1000); // update 1s from now
-        return;
-    }
+function getId ($child) {
+    let $d = $child.parentNode
+    while ($d && !$d.classList.contains('download')) $d = $d.parentNode
```
aidanholm: Same as `bookmarks_chrome`; prefer a do-while loop

lib/binds_chrome.lua
```diff
@@ -210,7 +207,7 @@ local mode_bind_template = [==[
     <li class="bind bind_type_{type}">
         <div class="link-box">
             <a href="#" class="filename">{filename}</a>
```
aidanholm: I think these two should be combined into one link that opens the editor. AFAIK the filename link does nothing.

In the link text, filename and line number should probably be separated by a colon, e.g: `./lib/binds.lua:164`

lib/bookmarks_chrome.lua
```diff
@@ -254,240 +254,247 @@ local html_template = [==[
...
+  const $next = document.getElementById('nav-next')
+  const $page = document.getElementById('page')
+  const $prev = document.getElementById('nav-prev')
+  const $results = document.getElementById('results')
+  const $search = document.getElementById('search')
+
+  $page.value = $page.value || 1
+
+  function makeBookmark (b) {
+    return createElement('div', { 'data-id': b.id, class: 'bookmark' }, [
```
aidanholm: This function does not correctly handle bookmarks that have no `tags` property.

lib/history_chrome.lua
```diff
@@ -117,210 +116,194 @@ local html_template = [==[
...
+
+  function updateClearButtons (all, results, selected) {
+    $clearAll.disabled = !!all
+    $clearResults.disabled = !!results
+    $clearSelected.disabled = !!selected
+  }
+
+  function updateNavButtons () {
+    $next.style.display = resultsLen === limit ? 'block' : 'none'
+    $prev.style.display = parseInt($age.value, 10) > 1 ? 'block' : 'none'
```
aidanholm: Should be `$page`, not `$age`

lib/history_chrome.lua
```diff
@@ -117,210 +116,194 @@ local html_template = [==[
...
+  const $next = document.getElementById('nav-next')
+  const $page = document.getElementById('page')
+  const $prev = document.getElementById('nav-prev')
+  const $results = document.getElementById('results')
+  const $search = document.getElementById('search')
+
+  $page.value = $page.value || 1
+
+  function makeHistoryItem (h) {
+    let domain = /:\/\/([^/]+)\//.exec(h.uri)
```
aidanholm: This function will break on `file://` URIs since no domain will be matched.

lib/downloads_chrome.lua
```diff
@@ -84,187 +84,136 @@ _M.stylesheet = [==[
...
-        }
+    if (downloads.length == null) {
+        setTimeout(update, 1000)
+        return
     }
+
+    downloads
+        .sort((a, b) => b.created - a.created)
+        .map(makeDownload)
+        .forEach($list.appendChild)
```
aidanholm, line 202: Should be `.forEach($d => $list.appendChild($d))`

lib/downloads_chrome.lua
```diff
@@ -84,187 +84,136 @@ _M.stylesheet = [==[
...
+                createElement('a', { href: uri }, [
+                    document.createTextNode(uri)
+                ])
+            ])
+
+        ]),
+
+        createElement('div', { class: 'controls' }, [
+
+            createElement('a', { class: 'show' }, [
```
aidanholm, line 164: These links all need `href=#`

aidanholm (commented):
Wow, thanks for all your work!

- Few minor changes and bug fixes in the diff comments.
- I'd prefer 4-space indentation, just to keep consistency with the rest of luakit.
- History next/prev buttons take up full page width for some reason?
- I'm not sure about the `createElement` approach. It makes the code harder to read and more difficult to track down bugs. I'd prefer a solution using tagged template literals.

aidanholm:
@taiite Any progress on this one?

ghost:
Hi, sorry I've been busy lately and haven't had the time to look at your comments. :(
I'm making changes now.

lib/binds_chrome.lua
ghost: I left the HTML structure unchanged and just put `data-` attributes instead of custom ones.

I'll push the fix in a moment.

lib/binds_chrome.lua
ghost: I'm not sure about that. jQuery documentation says that putting events on `body` and then filtering with CSS selectors is more time-consuming than putting events on the targeted elements directly.

That's the reason I used the `createElement` function, so that events are directly attached to the right element.

Source: http://api.jquery.com/on/#Event_performance

lib/binds_chrome.lua
ghost: I haven't thought about it much and just copy-pasted the former code. Modifying now, will push later. :)

lib/bookmarks_chrome.lua
ghost: Okay.

lib/bookmarks_chrome.lua
ghost: It was because `b.tags` is not a string when the bookmark doesn't have a tag. Fixed it.

lib/bookmarks_chrome.lua
ghost: Done

lib/bookmarks_chrome.lua
ghost: The former code also checked for errors at this point, so I put the try-catch block in the new version too.

lib/downloads_chrome.lua
ghost: `a` tags without an `href` attribute are OK in HTML5 though.

Source: https://stackoverflow.com/questions/5292343/is-an-anchor-tag-without-the-href-attribute-safe/19167175#19167175

lib/history_chrome.lua
ghost: This regex is the same as in the former code. I haven't much time today to find a fix to this one. :(

lib/history_chrome.lua
ghost: Thanks, I'll fix this typo.

lib/binds_chrome.lua
ghost: Changed it.

lib/downloads_chrome.lua
ghost: Done

ghost:
I've committed fixes for some of your reviews (though Github says all of them are outdated) and commented on most of them (the ones I haven't commented on are ones I haven't had the time to look at).

lib/history_chrome.lua
aidanholm: The regex itself should explicitly have `https?` at the start, and the case where it doesn't match (`domain` being `undefined`?) should be explicitly handled in the body of the function.

lib/downloads_chrome.lua
aidanholm: True, but placeholder links aren't desirable; e.g. without `href=#` they do not respond to clicks.

lib/binds_chrome.lua
aidanholm: That's a good point. That documentation refers to the performance of jQuery, however, which isn't necessarily valid for vanilla JS. As there'd only be a single event listener and a handful of `matches()` calls, there should be next-to-no performance difference.

lib/binds_chrome.lua
aidanholm: I noticed the `data-` fix, that's great!

aidanholm:
No worries! I've been busy myself

I'll take a look

lib/binds_chrome.lua
```diff
@@ -177,6 +173,10 @@ local html_template = [==[
         #templates {
             display: none;
         }
+
```
aidanholm, line 176: `.hidden` is already in `lib/chrome.lua`'s stylesheet; no need to duplicate it

lib/downloads_chrome.lua
ghost: Clicking on them works fine.
But I added the `href="#"` anyway.

lib/history_chrome.lua
ghost: Fixed!

lib/binds_chrome.lua
ghost: Okay it's been changed.

ghost:
Fixed some more reviews.

lib/downloads_chrome.lua
aidanholm: Hm, then I'm not sure what was causing that; not to worry, I guess :)

aidanholm:
I'll pull this one and do some work on it myself, then manually merge it

aidanholm:
This one's merged; thanks for your work! :)

PR #541, opened 2017-09-24
Make del_word in readline.lua UTF8-aware (fixes #525)

### LINKED ISSUE #525: `del_word` in `readline.lua` not UTF8-aware
The `del_word` function in `readline.lua` does not work as expected when the current contents of the line contain UTF8/wide characters. The reason is that `del_word` treats the position of the editing caret (which counts actually visible characters) directly as a character index to `string.sub` (which counts wide characters as several bytes/characters). To reproduce, use a UTF8 locale and enter ":äbc äbc äbc äbc", press `^W` a few times.

Looking at `del_word`, I would suppose that fixing this requires `string.wlen` (luakit's UTF8-aware `string.len` function), something like lua 5.3's `utf8.offset`, and/or UTF8-compatible `string.find` functions.

I can offer a PR that implements a `string.woffset` function (like `utf8.offset` in lua 5.3) using glib, and a modified `del_word` function (that currently considers all wide characters as non-word characters -- not sure how to best fix that). Please let me know if this could be helpful, or if there is a better way to do this.

hofheinz:


hofheinz:
This is a very crude attempt at a UTF8 library stub (only `utf8.len` and `utf8.offset` currently exist, but have the same interface as in Lua 5.3).

Most of the content is copied/pasted from other `common/clib/*.c` files, but my understanding of the Lua/C interaction is very limited. Hence, I guess there are still a number of problems (which I'd be happy to try to fix if made aware).

common/clib/utf8.c
```diff
@@ -0,0 +1,116 @@
...
+    return 1;
+}
+
+/* UTF8 aware string length computing.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_len(lua_State *L)
+{
+    const gchar *cmd = luaL_checkstring(L, 1);
+    lua_pushnumber(L, (ssize_t) g_utf8_strlen(NONULL(cmd), -1));
```
aidanholm: `NONULL` not necessary here; checkstring will always return a non-null pointer

common/clib/utf8.c
```diff
@@ -0,0 +1,116 @@
...
+luaH_utf8_new(lua_State *L)
+{
+    luaH_class_new(L, &utf8_class);
+    return 1;
+}
+
+/* UTF8 aware string length computing.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_len(lua_State *L)
```
aidanholm, line 42: `utf8.len` needs to support invalid strings: "If it finds any invalid byte sequence, returns a false value plus the position of the first invalid byte." You can use `g_utf8_validate` for this.

common/clib/utf8.c
```diff
@@ -0,0 +1,116 @@
...
+
+/* UTF8 aware string offset conversion.
+ * Converts (1-based) UTF8 offset to (1-based) byte offset.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_offset(lua_State *L)
+{
+    const gchar *cmd = luaL_checkstring(L, 1);
+    gint widx = luaL_checkint(L, 2);
+    const gint len = g_utf8_strlen(NONULL(cmd), -1);
```
aidanholm: Ditto re `NONULL`

common/clib/utf8.c
```diff
@@ -0,0 +1,116 @@
...
+    gint ret = 0;
+
+    /* convert negative to positive index */
+    if(widx < 0 && -widx <= len)
+        widx = len + widx + 1;
+
+    /* convert positive UTF8 offset to byte offset */
+    if(widx > 0 && widx <= len + 1) {
+        gchar *pos = g_utf8_offset_to_pointer(NONULL(cmd), widx - 1);
```
aidanholm: Ditto re `NONULL`

common/clib/utf8.c
aidanholm: `utf8.len` also takes two optional integer arguments representing a byte slice of the string

common/clib/utf8.c
```diff
@@ -0,0 +1,116 @@
...
+    /* if conversion was successful, output result (else output nil) */
+    if(ret > 0)
+        lua_pushnumber(L, (ssize_t) ret);
+    else
+        lua_pushnil(L);
+    return 1;
+}
+
+void
+utf8_class_setup(lua_State *L)
```
aidanholm: You've defined `utf8` as a class, which means that Lua code can create instances of it; it should be a library instead :)  See `clib/msg.c` for a minimal library to refer to.

common/clib/utf8.c
```diff
@@ -0,0 +1,116 @@
...
+    const gchar *cmd = luaL_checkstring(L, 1);
+    lua_pushnumber(L, (ssize_t) g_utf8_strlen(NONULL(cmd), -1));
+    return 1;
+}
+
+/* UTF8 aware string offset conversion.
+ * Converts (1-based) UTF8 offset to (1-based) byte offset.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_offset(lua_State *L)
```
aidanholm, line 75: `utf8.offset` also takes an optional `i` parameter

aidanholm (changes requested):
This looks great! I only have a few minor comments; the biggest one is the class vs library issue

hofheinz:
Thanks for the detailed review! (Sorry, I completely overlooked the parameters, and did not know any better than to make an instantiable class. Hopefully fixed. But, if I may say so, converting between 0-based [glib] and 1-based [lua] byte and UTF8 offsets really is a pain.)

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "common/clib/utf8.h"
+#include "luah.h"
+
+#include <glib.h>
+
+static lua_class_t utf8_class;
```
aidanholm: This isn't needed for a lib

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+LUA_CLASS_FUNCS(utf8, utf8_class)
+
+lua_class_t *
```
aidanholm: ditto

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+    return offset;
+}
+
+/* UTF8 aware string length computing.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_len(lua_State *L)
+{
+    const int nargs = lua_gettop(L);
+    const gchar *str = luaL_checkstring(L, 1);
```
aidanholm: Use `luaL_checklstring()` instead of a separate `strlen()` call

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+    ssize_t blen = strlen(str);
+    ssize_t bbeg = 0;
+    ssize_t bend = blen;
+    gchar *valend;
+
+    /* parse optional begin/end parameters */
+    if(nargs > 1) {
```
aidanholm: Use `luaL_optinteger()`

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+lua_class_t *
+utf8_lib_get_utf8_class(void)
+{
+    return &utf8_class;
+}
+
+/* Convert 1-based into 0-based byte offset,
+ * counted from back of string if negative */
+static ssize_t
+luaH_utf8_convert_offset(ssize_t offset, ssize_t length) {
```
aidanholm: I'd prefer this be exactly compatible with Lua5.3: https://github.com/lua/lua/blob/b1daa069bac27055f17753e85dca6aa256efef6f/lutf8lib.c#L30

This means that `offset > length` is an error condition

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+    const ssize_t blen = strlen(str);
+    ssize_t widx = luaL_checkint(L, 2);
+    if(widx > 0) widx--; /* adjust to 0-based */
+    ssize_t bbase = (widx>=0) ? 0 : blen;
+    ssize_t ret = 0;
+    ssize_t bbeg = 0;
+    ssize_t wseglen;
+
+    /* parse optional parameter (base index) */
+    if(nargs > 2)
```
aidanholm: `luaL_optinteger`

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/* UTF8 aware string offset conversion.
+ * Converts (1-based) UTF8 offset to (1-based) byte offset.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_offset(lua_State *L)
+{
+    const int nargs = lua_gettop(L);
+    const gchar *str = luaL_checkstring(L, 1);
```
aidanholm: `luaL_checklstring`

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+    const ssize_t blen = strlen(str);
+    ssize_t widx = luaL_checkint(L, 2);
```
aidanholm: `luaL_checkinteger`

common/clib/utf8.c
```diff
@@ -0,0 +1,141 @@
...
+    /* convert positive UTF8 offset to byte offset */
+    if(widx >= 0 && widx <= wseglen) {
+        gchar *pos = g_utf8_offset_to_pointer(str + bbeg, widx);
+        if (pos != NULL)
+            ret = (ssize_t) (pos - str) + 1;
+    }
+
+    /* if conversion was successful, output result (else output nil) */
+    if(ret > 0)
+        lua_pushnumber(L, ret);
```
aidanholm: `lua_pushinteger`

tests/async/test_clib_utf8.lua
```diff
@@ -0,0 +1,34 @@
...
+local assert = require "luassert"
+
+local T = {}
+
+T.test_module = function ()
+    assert.is_table(utf8)
+end
+
+T.test_utf8_len = function ()
+    assert(utf8.len("")==0)
```
aidanholm: Test cases look good! But these should be using `assert.equal()`; first argument should be the number, e.g:
```
assert.equal(1, utf8.len("ä"))
```

aidanholm (changes requested):

hofheinz:
Thanks for the constructive comments! I tried to mimic Lua 5.3's
implementation better. There is one corner case that I find hard to imitate
(without introducing ugliness): `uf8.len("ä",1,1)` (intuitively, the UTF8
length of the byte-substring [1,1] of `"ä"` -- note that since `"ä"` is wide,
this byte-substring is not valid UTF8). The PR code returns an error (because
the byte-substring is not valid UTF8), but Lua 5.3 returns `1`. (**Edit:** the PR code
actually returns `nil` plus an offset, as suggested in the `utf8.len` Lua manual:
"If it finds any invalid byte sequence, returns a false value plus the position of the
first invalid byte.")

Looking at Lua's source code (from the file you cited), the specification is:
```
utf8len(s [, i [, j]]) --> number of characters that start in the
range [i,j], or nil + current position if 's' is not well formed in
that interval
```
To me, this doesn't sound completely define what should happen with `utf8len("ä",1,1)`. Also, it seems to me that the obvious way to get Lua's behavior is to backtrack from position `j` (or to move forward from position `i` until we're at or beyond `j`) and check individual UTF8 characters. (This is basically what the Lua code does. But in our situation, since we already have `glib` functions, I find this a bit ugly.)

What should we do here? Have an inconsistency with Lua 5.3 (in a somewhat exotic corner case), or imitate Lua 5.3 at the cost of a cumbersome implementation?

aidanholm:
The spec comment you cite says "number of characters that _start_ in the range [i,j]. For that example, the range covers the first byte of "ä", so there is no inconsistency there in the 5.3 code/spec. The docs are a little ambiguous regarding the error condition; they should read "If any invalid byte sequence _beginning in the range specified by [i,j]_ is found"

I'd prefer to be compatible with Lua 5.3 here. You could use `g_utf8_find_next_char` on the last byte in the range [i,j] to increase `bend` to include any partial character at the end of the string. Then validate the possibly expanded range with `g_utf8_validate`. This should fix both the incorrect counting of ä, and the error handling for invalid bytes at the end of the string.

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+        return -1;
+    return offset;
+}
+
+/* UTF8 aware string length computing.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_len(lua_State *L)
+{
+    ssize_t blen;
```
aidanholm: Should be `size_t`, not `ssize_t`

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+    const gchar *str = luaL_checklstring(L, 1, &blen);
+    gchar *valend;
+    ssize_t bbeg, bend;
```
aidanholm: ditto

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+
+    /* parse optional begin/end parameters
+     * imitate Lua 5.3: raise an error if out of bounds
+     */
+    bbeg = luaL_optinteger(L, 2, 1);
+    bbeg = luaH_utf8_convert_offset(bbeg, blen);
+    luaL_argcheck(L, bbeg != -1, 2, "initial position out of string");
+    bend = luaL_optinteger(L, 3, blen) + 1;
+    bend = luaH_utf8_convert_offset(bend, blen);
+    luaL_argcheck(L, bend != -1, 3, "final position out of string");
```
aidanholm: `bend < bbeg` is also an error condition in 5.3

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+    if(bend < bbeg)
+        bend = blen;
+
+    /* is the string valid UTF8? */
+    if(!g_utf8_validate(str + bbeg, bend - bbeg, (const gchar **) &valend)) {
```
aidanholm: `valend` should be defined where it is used

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+    return 1;
+}
+
+/* UTF8 aware string offset conversion.
+ * Converts (1-based) UTF8 offset to (1-based) byte offset.
+ * Returns the number of elements pushed on the stack. */
+static gint
+luaH_utf8_offset(lua_State *L)
+{
+    ssize_t blen;
```
aidanholm: `size_t`

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+#include "luah.h"
+
+#include <glib.h>
+
+/* Convert 1-based into 0-based byte offset,
+ * counted from back of string if negative
+ * return -1 if offset is out of range */
+static ssize_t
+luaH_utf8_convert_offset(ssize_t offset, ssize_t length) {
+    if(offset > 0)
```
aidanholm: There should be a space between `if` and `(`

doc/luadoc/utf8.lua
```diff
@@ -0,0 +1,18 @@
+--- Basic UTF8 character counting support for Luakit.
+--
+-- @class utf8
+-- @author Dennis Hofheinz
+-- @copyright 2017 Dennis Hofheinz <github@kjdf.de>
+
+--- @method len
+-- Return the length (in UTF8 characters) of a string.
+-- @tparam string string The string whose length is to be returned.
```
aidanholm: This should include:
- information on the optional parameters
- longer description of how the function works, error conditions
- a "compatible with lua 5.3 utf8.len" comment

doc/luadoc/utf8.lua
```diff
@@ -0,0 +1,18 @@
...
+-- @treturn integer The length of that string.
+
+--- @method offset
+-- Convert an offset (in UTF8 characters) to a byte offset.
```
aidanholm: ditto

tests/async/test_clib_utf8.lua
```diff
@@ -0,0 +1,34 @@
...
+end
+
+T.test_utf8_len = function ()
+    assert.equal(0, utf8.len(""))
+    assert.equal(1, utf8.len("ä"))
+    assert.equal(2, utf8.len("äa"))
+    assert.equal(1, utf8.len("äa", -1))
+    assert.equal(2, utf8.len("äa", -3))
+    assert.equal(1, utf8.len("äa", 1, 2))
+    assert.equal(2, utf8.len("äa", 1, 3))
```
aidanholm: Looks good! Will also need the `utf8.len("ä", 1, 1)` test case here too

aidanholm (changes requested):

common/clib/utf8.c
hofheinz: `bbeg` and `bend` are currently signed because `luaH_utf8_convert_offset` may return a `-1` as an error indicator. This means having `blen` should also be `ssize_t` because we compare `bend` with `blen` at some point. Alternative: explicit type casts, or have `luaH_utf8_convert_offset` do something else in case of out-of-bounds errors. Not sure what to do (but I do like the concept of a helper function that converts negative to positive offsets and does out-of-bounds checks).

common/clib/utf8.c
hofheinz: I couldn't reproduce this. Running `lua53 -e 'print(utf8.len("abc",2,1))'` yields `0`. (`lua53 -v` yields `Lua 5.3.4  Copyright (C) 1994-2017 Lua.org, PUC-Rio`, this seems up to date.) Do you get different results?

aidanholm: Sorry, I misread the Lua code for `utflen()`; the actual correct behavior is to return 0 if j < i, but I think your code handles this correctly already.

common/clib/utf8.c
aidanholm: Unsigned overflow is well-defined; just return and compare against `(size_t)-1`. The other alternative is to pass in `L` and throw an error from the helper function, but that seems messier.

common/clib/utf8.c
```diff
@@ -0,0 +1,134 @@
...
+    ssize_t blen;
+    const gchar *str = luaL_checklstring(L, 1, &blen);
+    gchar *valend;
+    ssize_t bbeg, bend;
+
+    /* parse optional begin/end parameters
+     * imitate Lua 5.3: raise an error if out of bounds
+     */
+    bbeg = luaL_optinteger(L, 2, 1);
+    bbeg = luaH_utf8_convert_offset(bbeg, blen);
```
aidanholm: I think the helper function should have a short name, so you can one-line these; something like `abspos`

hofheinz:
Thanks for the hints! (Your `g_utf8_find_next_char` suggestion worked nicely!) This is not finished yet, though, one of the corner cases tests is still failing. (Will get back to this.)

hofheinz:
Ok, tests pass now.

aidanholm:
Looks pretty good! I'll pull this, do some last fixups, and then manually merge it

aidanholm:
Merged at 3692d295; thanks for your work!

hofheinz:
Great! Thanks for your comments, I learned a lot.

aidanholm:
Glad to hear :)

Issue #542, opened 2017-09-25
Latest Release - No Longer Loading in Hi-Dpi mode with zoom change in globals.lua


IWriteThings:
What do I need to change in the rc.lua file to load in Hi-Dpi mode? The zoom method is no longer working on update.

aidanholm:
Try adding `settings.webview.zoom_level = 200` to your `rc.lua`, or changing the value `webview.zoom_level` on the settings page (`luakit://settings/`)

IWriteThings:
Adding `settings.webview.zoom_level = 200` to the `rc.lua` did not work.

However, changing the webview.zoom_level under the settings page functioned perfectly for me. Thanks a bunch! I am going to open a pull request to update the readme with this information for other users.

IWriteThings:
I updated the readme file with a configuration update to help others who may run into this issue.

Thanks a bunch!
Aaron

aidanholm:
Adding `settings.webview.zoom_level = 200` to `rc.lua` should work; was there any error message?

Thanks for the PR! I've added some minor requests
 \- Aidan

IWriteThings:
There was no error message. It simply did not change the size of the text. I added the above text to my rc.lua and then I restarted the browser by closing it and reopening it.

aidanholm:
Ah, I'm guessing you added it at the very end of `rc.lua`? You have to add it before the `End user script loading` comment, as the code after that opens the initial windows/tabs, and the setting is only reapplied after a page reload.

PR #545, opened 2017-09-25
Update README with config for HiDPI

IWriteThings:
Adds configuration instructions for HiDPI monitors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,9 @@ use those files when you next launch it.
 
 ## HiDPI Monitor Configuration
 
-If you have a HiDPI monitor (> 1920x1080p) and need to increase the size of the font loaded, you can do so by changing the webview.zoom_level on the settings page (luakit://settings/) to 150 or 200 as per your taste.
+If you have a HiDPI monitor (> 1920x1080) and find that web pages are too small,
+you can change the `webview.zoom_level` on the settings page (luakit://settings/)
+to 150 or 200 as per your taste.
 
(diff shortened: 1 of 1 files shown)


IWriteThings:
This would close ticket #542 . Please review this change for acceptability.

README.md
```diff
@@ -107,6 +107,10 @@ Just copy the files you wish to change (and the rc.lua) into
 `$XDG_CONFIG_HOME/luakit` (defaults to `~/.config/luakit/`) and luakit will
 use those files when you next launch it.
 
+## HiDPI Monitor Configuration
+
+If you have a HiDPI monitor (> 1920x1080p) and need to increase the size of the font loaded, you can do so by changing the webview.zoom_level on the settings page (luakit://settings/) to 150 or 200 as per your taste.
```
aidanholm: - I'd prefer if this long line were wrapped to a reasonable column length.
- `1920x1080p` -> `1920x1080`
- "and need to increase the size of the font loaded" -> "and find that web pages are too small"; technically this setting changes more than just the font size
- `webview.zoom_level` should be surrounded with backticks so that it renders in monospace.

aidanholm (changes requested):
Thanks for the PR! Just a few minor suggestions really

IWriteThings:
I have updated the commits. Can you please review? I believe this meets all requests.

aidanholm:
Merged! Thanks :)

PR #547, opened 2017-09-26
Remove LIBDIR on uninstall target

ghost:


aidanholm:
Oops, guess I missed that one :) thanks!

PR #549, opened 2017-10-02
Correct substituted buildopts.h path strings

vlnx:
Allows compiling in a work directory and setting the buildopts.h.in paths to the final installed path
Disables relative development paths while keeping DEVELOPMENT_PATHS=1 working


aidanholm:
Thanks for the PR! This looks pretty good; I'll be busy for a few days but will give it a review when I get some free time

aidanholm:
I've been testing this and one issue I've discovered is that it breaks tests, which rely on `luakit.so` being in the current directory. I'm trying to come up with a better way of implementing that.

c0dev0id:
I think this has been superseded by another PR that fixed DESTDIR. I'll close this 7 year old PR. But feel free to reopen an rebase it on current luakit.

PR #563, opened 2017-10-16
remove_bind returns the removed binding

arcnmx:
Many modules implement their bindings using anonymous functions, so this
provides a way to remove an existing key or command and then re-bind it
using a different combination.

The return type is a table including the arguments in the order originally
passed to `add_bind`, though since the input isn't a table would it be more
appropriate to use `unpack()` here to use a multiple return instead?

I'd also like to add `modes.rebind("mode", "<new>", "<old>", keep_old)` in
another PR if there's interest in providing this with a more convenient API
built on top of this change. If not, then at least this enables an easier
implementation inside of `rc.lua`.

### DIFF
--- a/lib/lousy/bind.lua
+++ b/lib/lousy/bind.lua
@@ -420,6 +420,9 @@ end
 --- Remove any binding with a specific trigger from the given array of bindings.
 -- @tparam table binds The array of bindings to remove the named binding from.
 -- @tparam string bind The trigger to unbind.
+-- @treturn table The associated action of the binding that was removed, or `nil` if
+-- not found.
+-- @treturn table The options of the binding that was removed, or `nil` if not found.
 function _M.remove_bind (binds, bind)
     assert(binds and type(binds) == "table", "invalid binds table type: " .. type(binds))
     assert(bind and type(bind) == "string", "invalid bind type: " .. type(bind))
@@ -428,10 +431,11 @@ function _M.remove_bind (binds, bind)
         if m[1] == bind then
             table.remove(binds, i)
             msg.verbose("removed bind %s", bind)
-            return
+            return m[2], m[3]
         end
     end
     msg.verbose("no bind %s to remove", bind)
+    return nil, nil
 end
 
 return _M


aidanholm:
This looks good, thanks! I agree that the return value should be the same as the input.

A `rebind()` API sounds great, would love that!

lib/lousy/bind.lua
```diff
@@ -420,18 +420,20 @@ end
...
     assert(binds and type(binds) == "table", "invalid binds table type: " .. type(binds))
     assert(bind and type(bind) == "string", "invalid bind type: " .. type(bind))
     bind = convert_bind_syntax(bind)
     for i, m in ipairs(binds) do
         if m[1] == bind then
-            table.remove(binds, i)
+            local binding = table.remove(binds, i)
             msg.verbose("removed bind %s", bind)
-            return
+            return binding
```
aidanholm: To return the action and opts, this should be `return m[2], m[3]`

lib/lousy/bind.lua
```diff
@@ -420,18 +420,20 @@ end
...
 -- @tparam string bind The trigger to unbind.
+-- @treturn table The binding that was removed, or nil if it could not be found.
 function _M.remove_bind (binds, bind)
     assert(binds and type(binds) == "table", "invalid binds table type: " .. type(binds))
     assert(bind and type(bind) == "string", "invalid bind type: " .. type(bind))
     bind = convert_bind_syntax(bind)
     for i, m in ipairs(binds) do
         if m[1] == bind then
-            table.remove(binds, i)
+            local binding = table.remove(binds, i)
```
aidanholm: No need to add another local var, as `m` is already set to this value.

arcnmx:
Changes made! Not entirely sure about the not found case, `return nil, nil` vs `return nil` vs `return` vs no return etc. are all functionally equivalent so just it's a matter of style.

lib/lousy/bind.lua
```diff
@@ -420,6 +420,9 @@ end
 --- Remove any binding with a specific trigger from the given array of bindings.
 -- @tparam table binds The array of bindings to remove the named binding from.
 -- @tparam string bind The trigger to unbind.
+-- @treturn table The associated action of the binding that was removed, or nil if
```
aidanholm: nil here should be surrounded in backticks: \`nil\`

lib/lousy/bind.lua
```diff
@@ -420,6 +420,9 @@ end
...
+-- not found.
+-- @treturn table The options of the binding that was removed, or nil if not found.
```
aidanholm: ditto

aidanholm (commented):
minor nit :)

aidanholm:
That's a common misconception; they actually _aren't_ equivalent. If you assign to multiple variables then any excess variables will be set to `nil`, but `return` literally means return no values!

aidanholm:
Last request: could you squash all these into a single commit? Thanks!

aidanholm:
Thanks :) all done!

hippi777:
hi there! :)

@aidanholm am i right, if i'm saying that the return vs return nil difference wont ever appear in the result (i couldnt produce anything visible), but it will make some extra steps, so omitting whatever can be omitted is just a best practise for microoptimalization? - or is there anything else?

just reread your msg, so the diff you are talking about that the nils in some multiple assigment wont come form the function, so its still only have anything to do with performance, but nothing else - maybe this one is right? :D 

and thx! :)

aidanholm:
@szbnwer The usual case where it matters is when calling Lua functions and signal handlers from C. All return values are pushed onto an internal stack, which the C code manipulates. There used to be bugs with the C signal handling implementation for this reason, which would stop signal handling if a signal handler ended in `return nil`, but not if it ended in `return`.

arcnmx:
Yeah, I meant functionally equivalent on the caller side in the context of how the language assigns to multiple values - FFI that doesn't follow the same rules certainly can be affected, and there are probably ways to observe the difference in pure lua using varargs.

hippi777:
fine, thx for both of you! :)

PR #564, opened 2017-10-18
modes.rebind()

arcnmx:
First pass at a `rebind()`. Consider this an RFC as there are a couple decisions to make.

1. There are a few ways to expose this publicly:
    1. Implement as `lousy.binds.rebind()` with a thin wrapper in `modes.rebind()`.
    2. Implement directly as `modes.rebind()`
2. Then there are two main alternatives to what the API should be:
    1. `modes.rebind("mode", "<new>", "<old>", keep)`
    2. `modes.rebind("mode", { { "<new>", "<old>", keep } })`
        - If `lousy.binds.rebind()` exists it will still use the first option.
3. There are also a few ways to actually implement it:
    1. Using `lousy.bind.remove_bind`, but slightly suboptimal due to `keep` requiring us to
re-add it.
        - Besides being inefficient, this subtly changes the order of the bind tables.
          Not sure if anything relies on that ordering.
    2. Loop through the bind table and just reassign `m[1]` (or append depending on `keep`)
4. Should it return a boolean indicating success?

### DIFF
--- a/lib/lousy/bind.lua
+++ b/lib/lousy/bind.lua
@@ -438,6 +438,34 @@ function _M.remove_bind (binds, bind)
     return nil, nil
 end
 
+--- Remap a binding from a given trigger to a new trigger, optionally keeping
+-- the original binding. In both cases, the new binding will have the same
+-- options as the original binding.
+-- @tparam table binds The array of bindings to remap within.
+-- @tparam string new The new trigger to map from.
+-- @tparam string old The existing trigger to remap.
+-- @tparam[opt] boolean keep Retain the existing binding.
+-- @default `false`
+function _M.remap_bind (binds, new, old, keep)
+    assert(binds and type(binds) == "table", "invalid binds table type: " .. type(binds))
+    assert(new and type(new) == "string", "invalid bind type: " .. type(new))
+    assert(old and type(old) == "string", "invalid bind type: " .. type(old))
+    new = convert_bind_syntax(new)
+    old = convert_bind_syntax(old)
+    for _, m in ipairs(binds) do
+        if m[1] == old then
+            msg.verbose("remapping bind to %s, %s %s", new, keep and "keeping" or "removing", old)
+            if keep then
+                _M.add_bind(binds, new, m[2], m[3])
+            else
+                m[1] = new
+            end
+            return
+        end
+    end
+    msg.verbose("no bind %s to remap", old)
+end
+
 return _M
 
 -- vim: et:sw=4:ts=8:sts=4:tw=80
--- a/lib/modes.lua
+++ b/lib/modes.lua
@@ -225,7 +225,7 @@ _M.new_mode("lua", [[Execute arbitrary Lua commands within the luakit
 --         end},
 --     })
 --
--- ## Bind format
+-- # Bind format
 --
 -- Every item in the `binds` array must be a table that defines a single binding
 -- between a trigger and an action. Each entry must have the following form:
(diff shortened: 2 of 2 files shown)


aidanholm:
1) Would prefer (i) as it's more in line with the rest of the binds/modes APIs. `lousy.bind` is for modifying bindsets, `modes` is for modifying the bindsets for a particular mode.
2) Would prefer (ii), again it's more consistent, even if it's a little ugly.
3) If `keep` is falsey, loop through and change the bind string of the first match. Otherwise use `add_bind()`.
4) No, as none of the others do.

Additionally, I'd prefer that the functions be named `remap_bind` and `remap_binds`, to match the others.

arcnmx:
Take 2!

1. Probably bikeshedding to be done on my choice of argument names and documentation wording.
2. Is `keep` the right polarity, or should it be `remove` instead? Is removal of the old binding the right default?
3. `msg.verbose()` print out once we're done? This is already done by `add_bind` though so... Only on false `keep`?

lib/lousy/bind.lua
```diff
@@ -438,6 +438,31 @@ function _M.remove_bind (binds, bind)
     return nil, nil
 end
 
+--- Remap an existing binding with a new trigger, optionally duplicating it.
```
aidanholm: "Remap a binding from a given trigger to a new trigger, optionally keeping the original binding. In both cases, the new binding will have the same options as the original binding."

lib/lousy/bind.lua
```diff
@@ -438,6 +438,31 @@ function _M.remove_bind (binds, bind)
...
+-- @tparam table binds The array of bindings to remap within.
+-- @tparam string new The new trigger to map to.
```
aidanholm: map from, not to

lib/lousy/bind.lua
```diff
@@ -438,6 +438,31 @@ function _M.remove_bind (binds, bind)
...
+-- @tparam string old The existing trigger to change.
```
aidanholm: remap, not change

lib/lousy/bind.lua
```diff
@@ -438,6 +438,31 @@ function _M.remove_bind (binds, bind)
...
+-- @tparam[opt] boolean keep Retain the existing binding.
```
aidanholm: 👍

lib/lousy/bind.lua
```diff
@@ -438,6 +438,31 @@ function _M.remove_bind (binds, bind)
...
+-- @default `false`
+function _M.remap_bind (binds, new, old, keep)
+    assert(binds and type(binds) == "table", "invalid binds table type: " .. type(binds))
+    assert(new and type(new) == "string", "invalid bind type: " .. type(new))
+    assert(old and type(old) == "string", "invalid bind type: " .. type(old))
+    new = convert_bind_syntax(new)
+    old = convert_bind_syntax(old)
+    for _, m in ipairs(binds) do
+        if m[1] == old then
```
aidanholm: Should add logging here, on the line before `if keep then`

lib/modes.lua
```diff
@@ -283,6 +283,40 @@ _M.remove_binds = function (mode, binds)
...
+--- Bind an existing key or command to a new binding.
+--
+-- # Example
+--
+--     -- Add an additional zooming command binding
+--     modes.rebind("normal", {
+--         { "<Control-=>", "zi", true },
+--     })
+--
+-- ## Bind format
```
aidanholm: #, not ##

lib/modes.lua
```diff
@@ -283,6 +283,40 @@ _M.remove_binds = function (mode, binds)
...
+--
+-- Every item in the `binds` array must be a table that defines a single rebind
+-- from an old to new trigger. Each entry must have the following form:
```
aidanholm: "from an existing trigger to a new one"

aidanholm (commented):
1) The code itself looks great! Just documentation fussiness from me :)

2) `keep` is the right polarity, as the default behavior should be to change the existing binding, not make a new one, and that's what you'll get if you leave off the `keep` parameter.

3) I'd say log a verbose message with the two trigger strings and whether the old one is being kept, at the point indicated

lib/modes.lua
arcnmx: Good catch, I stole it from [here](https://github.com/luakit/luakit/blob/develop/lib/modes.lua#L228) so you may want to do a quick fix on that one too.

arcnmx:
Feel free to suggest better wording for the log message.

lib/modes.lua
aidanholm: Fixed, thanks!

lib/lousy/bind.lua
```diff
@@ -438,6 +438,34 @@ function _M.remove_bind (binds, bind)
...
+            msg.verbose("remapping bind %s, %s %s", new, keep and "keeping" or "removing", old)
```
aidanholm: Final nit: should be "remapping bind _to_ %s"

aidanholm:
Looks good; final request is to squash all commits into one, with the commit message "Add lousy.bind.remap_bind(), modes.remap_binds()"

aidanholm:
Merged! Thanks :)

PR #565, opened 2017-10-20
Fix missing line from modes.remap_binds()

arcnmx:
I have no idea how this managed to disappear from the function :(


aidanholm:
Good catch! I completely missed this one - thanks :)

PR #597, opened 2017-11-16
Add function to simulate keystrokes.

amosbird:
This patch is a rebase of the keysym branches, with some improvements and tenderly care of @aidanholm :)


lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
+--- Keysym for luakit.
```
aidanholm: I'm not happy with the name; should be "sendkeys" or "fakekeys" or something, but "keysym" is totally opaque

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
...
+-- This module parses a vim-like keystring into single keys and sends
+-- them to the window. A keystring is a string of keys to press, with
+-- special keys denoted in between angle brackets:
+--
+--  <code>keysym.send(w, "<S-Home><BackSpace>")</code>
+--
+-- See gdk/gdkkeysyms.h for a complete list of recognized key names.
+--
+-- @module keysym
+-- Simulating keystrokes
```
aidanholm: remove extra line "Simulating keystrokes" here

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
+--- Keysym for luakit.
+--
+-- This module parses a vim-like keystring into single keys and sends
+-- them to the window. A keystring is a string of keys to press, with
+-- special keys denoted in between angle brackets:
+--
+--  <code>keysym.send(w, "<S-Home><BackSpace>")</code>
```
aidanholm: indent code with four spaces instead of using `<code>`

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
...
+    local symbol = nil
+    local modifiers = {}
+    local keys = {}
+    -- https://www.lua.org/manual/5.3/manual.html#pdf-utf8.codes
+    for char in string.gmatch(keystring, "[%z\1-\x7F\xC2-\xF4][\x80-\xBF]*") do
+        if char == "<" then
+            symbol = ""
+        elseif char == ">" and symbol then
+            if #symbol == 0 then
+                w:error("bad keystring: " .. keystring)
```
aidanholm: - these are script errors, not user-facing error messages: use `error()` instead
- error message needs to say what's wrong with the keystring

a better version would be `error("keystring '" .. keystring .. "' contains empty modifier group")`; note the string concatenation; unfortunately `error()` doesn't support format strings (might hack that in at some point..)

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
...
+                })
+            end
+            symbol = nil
+            modifiers = {}
+        elseif symbol and char == "-" then
+            if string.match(symbol, "^[SCL]$") or string.match(symbol, "^M[1-5]$") then
+                table.insert(modifiers, symbol)
+                symbol = ""
+            else
+                w:error("bad modifier in keystring: " .. symbol)
```
aidanholm: ditto

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
...
+            table.insert(keys, {
+                key = char,
+                mods = {},
+            })
+        else
+            symbol = symbol .. char
+        end
+    end
+    if symbol then
+        w:error("bad keystring: " .. keystring)
```
aidanholm: ditto

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
...
+        return
+    end
+    for _, key in ipairs(keys) do
+        if not w.win:send_key(key.key, key.mods) then
+            w:error("failed to send keystroke: " .. key.key)
```
aidanholm: ditto

lib/window.lua
```diff
@@ -166,7 +166,11 @@ local init_funcs = {
     end,
 
     key_press_match = function (w)
-        w.win:add_signal("key-press", function (_, mods, key)
+        w.win:add_signal("key-press", function (_, mods, key, send)
```
aidanholm: `send` should be `synthetic`

lib/window.lua
```diff
@@ -166,7 +166,11 @@ local init_funcs = {
...
+            -- avoid double mapping
+            if send then
```
aidanholm: this should be all one line; but the real problem is that this breaks people using external tools to automate luakit (is this likely? no, but it's still an API break)

I think the best solution for now is to add a `window.act_on_synthetic_keys` setting (feel free to think of a nicer name), and add it to the if-condition.

widgets/common.c
```diff
@@ -33,7 +33,8 @@ key_press_cb(GtkWidget* UNUSED(win), GdkEventKey *ev, widget_t *w)
     luaH_object_push(L, w->ref);
     luaH_modifier_table_push(L, ev->state);
     luaH_keystr_push(L, ev->keyval);
-    gint ret = luaH_object_emit_signal(L, -3, "key-press", 2, 1);
+    lua_pushboolean(L, ev->send_event);
```
aidanholm, line 36: :+1:

widgets/window.c
```diff
@@ -71,6 +72,83 @@ luaH_window_set_default_size(lua_State *L)
...
+    if (!keyval || keyval == GDK_KEY_VoidSymbol) {
+        lua_pushboolean(L, FALSE);
+        return 1;
+    }
+
+    guint state = 0;
+    luaH_checktable(L, 3);
+
+    lua_pushnil(L);
+    while(lua_next(L, 3)) {
```
aidanholm: I'm not happy with this bind syntax being different to the bind syntax used for keybindings. It might be a bit of work unifying that, but I'd also rather not get locked into having two syntaxes for compatibility reasons...

widgets/window.c
```diff
@@ -71,6 +72,83 @@ luaH_window_set_default_size(lua_State *L)
...
+
+#undef MODKEY
+
+        lua_pop(L, 1);
+    }
+
+    GdkKeymapKey* keys = NULL;
+    gint n_keys;
+
+    if (!gdk_keymap_get_entries_for_keyval(gdk_keymap_get_default(), keyval, &keys, &n_keys)) {
```
aidanholm: `gdk_keymap_get_default()` should be `NULL`

widgets/window.c
```diff
@@ -71,6 +72,83 @@ luaH_window_set_default_size(lua_State *L)
...
+        lua_pushboolean(L, FALSE);
+        return 1;
+    }
+
+    g_assert(keys);
```
aidanholm: can probably get rid of this as well

widgets/window.c
```diff
@@ -71,6 +72,83 @@ luaH_window_set_default_size(lua_State *L)
...
+    event_key->send_event = TRUE;
+    event_key->time = GDK_CURRENT_TIME;
+    event_key->state = state;
+    event_key->keyval = keyval;
+    event_key->hardware_keycode = keys[0].keycode;
+    event_key->group = keys[0].group;
+
+    g_free(keys);
+
+    gdk_event_put(event);
```
aidanholm: I'd say move this `gdk_event_put()` call up against the `event_key` block above

widgets/window.c
```diff
@@ -71,6 +72,83 @@ luaH_window_set_default_size(lua_State *L)
 }
 
 static gint
+luaH_window_send_key(lua_State *L)
```
aidanholm: I haven't really looked at this code, but it's spamming warnings and that's a no no
```
Gdk-WARNING **: Event with type 8 not holding a GdkDevice. It is most likely synthesized outside Gdk/GTK+
```

aidanholm: Also, in general this could do with some logging: probably just a `debug("sending key '%s' to window %p", ...)` is sufficient. And the validation should use `luaL_error()` on failure: this is a function that only fails if the user screwed up the argument.

config/rc.lua
```diff
@@ -61,6 +61,9 @@ require "settings_chrome"
 local adblock = require "adblock"
 local adblock_chrome = require "adblock_chrome"
 
+-- Add ability to send keystrokes to the window
```
aidanholm: Do we want this enabled by default? I'm undecided...

lib/keysym.lua
```diff
@@ -0,0 +1,71 @@
...
+-- @copyright 2010 Mason Larobina mason.larobina@gmail.com
+
+local _M = {}
+
+_M.send = function (w, keystring)
+    local symbol = nil
+    local modifiers = {}
+    local keys = {}
+    -- https://www.lua.org/manual/5.3/manual.html#pdf-utf8.codes
+    for char in string.gmatch(keystring, "[%z\1-\x7F\xC2-\xF4][\x80-\xBF]*") do
```
aidanholm: as mentioned on IRC, ugh; I'll have a look at getting `utf8.codes` in

aidanholm (changes requested):

aidanholm:
In addition, there are two test failures:
```
FAIL style/test_source_format / test_vim_modeline
  tests/style/test_source_format.lua:41: Some files do not have modelines:
  lib/keysym.lua                                         Missing/malformed modeline
FAIL style/test_source_format / test_lua_module_functions_are_documented
  tests/style/test_source_format.lua:262: Some Lua modules have documentation issues:
  lib/keysym.lua:22                                      Undocumented export 'send'
```
The modeline is the `-- vim: et:sw=4:ts=8:sts=4:tw=80` thing that every file has. The new window `send_keys` method will need documenting too, but that's more of a final task.

aidanholm:
Oh, and another thing I'd prefer is if you could add your changes on top of the existing commits in the keysym branch, as that preserves history: both the reasoning behind each section of code, and the attribution to the original authors

widgets/window.c
amosbird: Doesn't work if I change it to `NULL`

lib/window.lua
```diff
@@ -709,6 +712,11 @@ function _M.ancestor(w)
 end
 
 settings.register_settings({
+    ["window.act_on_synthetic_keys"] = {
+        type = "boolean",
+        default = false,
+        desc = "Whether or not to act on syntheic keys.",
```
aidanholm: At the moment this is just restating the setting name; it should be more descriptive, e.g: "Whether synthetic key events should activate key bindings. Synthetic key events have been generated by a program rather than a physical key press, such as those sent by `keysym.send()`."

lib/keysym.lua
```diff
@@ -0,0 +1,72 @@
...
+local _M = {}
+
+--- Send synthetic keys to given window.
+-- @tparam w The window object.
+-- @tparam string keystring The key string representing synthetic keys.
+_M.send = function (w, keystring)
+    local symbol = nil
+    local modifiers = {}
+    local keys = {}
+    for char in string.gmatch(keystring, utf8.charpattern) do
```
aidanholm: The usual style here is to use method access notation: `keystring:gmatch(utf8.charpattern)`

lib/keysym.lua
```diff
@@ -0,0 +1,72 @@
...
+-- @copyright 2010 Fabian Streitel luakit@rottenrei.be
+-- @copyright 2010 Mason Larobina mason.larobina@gmail.com
+
+local _M = {}
+
+--- Send synthetic keys to given window.
+-- @tparam w The window object.
+-- @tparam string keystring The key string representing synthetic keys.
+_M.send = function (w, keystring)
+    local symbol = nil
```
aidanholm: this needs assertions that `w` is a table and `keystring` is a string

lib/keysym.lua
```diff
@@ -0,0 +1,72 @@
...
+            else
+                table.insert(keys, {
+                    key = symbol,
+                    mods = modifiers,
+                })
+            end
+            symbol = nil
+            modifiers = {}
+        elseif symbol and char == "-" then
+            if string.match(symbol, "^[Ss]hift$") or
```
aidanholm: ditto

lib/keysym.lua
```diff
@@ -0,0 +1,72 @@
...
+        elseif not symbol then
+            table.insert(keys, {
+                key = char,
+                mods = {},
+            })
+        else
+            symbol = symbol .. char
+        end
+    end
+    assert(not symbol)
```
aidanholm: This needs an "unterminated symbol" message

widgets/window.c
```diff
@@ -90,8 +89,7 @@ luaH_window_send_key(lua_State *L)
     }
 
     if (!keyval || keyval == GDK_KEY_VoidSymbol) {
```
aidanholm: No need for braces with a single line body

widgets/window.c
```diff
@@ -70,6 +71,78 @@ luaH_window_set_default_size(lua_State *L)
...
+    } else {
+        keyval = gdk_keyval_from_name(key_name);
+    }
+
+    if (!keyval || keyval == GDK_KEY_VoidSymbol) {
+        return luaL_error(L, "failed to get a valid key value.");
+    }
+
+    guint state = 0;
+    luaH_checktable(L, 3);
```
aidanholm: The checktable call should be moved to the top with the other argument checks

widgets/window.c
```diff
@@ -70,6 +71,78 @@ luaH_window_set_default_size(lua_State *L)
...
+    GdkEventKey *event_key = (GdkEventKey *) event;
+    event_key->window = gtk_widget_get_window(w->widget);
+    event_key->send_event = TRUE;
+    event_key->time = GDK_CURRENT_TIME;
+    event_key->state = state;
+    event_key->keyval = keyval;
+    event_key->hardware_keycode = keys[0].keycode;
+    event_key->group = keys[0].group;
+    gdk_event_put(event);
+    debug("sending key '%s' with modifier %d to window %p",
```
aidanholm: The window pointer printed should be the widget (`w`)

aidanholm (commented):
Just a few minor fixes needed

widgets/window.c
aidanholm: Actually, the table should be optional, but if there is a third argument it must be a table

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
     return 0;
 }
 
+static gint
+luaH_window_send_key(lua_State *L)
+{
+    widget_t *w = luaH_checkwidget(L, 1);
+    const gchar *key_name = luaL_checkstring(L, 2);
+    luaH_checktable(L, 3);
```
aidanholm: This needs to be inside `if (lua_gettop(L) >= 3)`, as we want to allow leaving the table out.

aidanholm: This needs to be inside `if (lua_gettop(L) >= 3)`, as we want to allow leaving the table out.

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
...
+
+    if (!g_utf8_validate(key_name, -1, NULL))
+        return luaL_error(L, "key name isn't a utf-8 string.");
```
aidanholm: no full stop on error messages

widgets/window.c
aidanholm: in order not to add another check below in the modifier loop, you could add an `else lua_newtable(L);` to that

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
...
+        MODKEY("mod5", MOD5);
+
+#undef MODKEY
+
+        lua_pop(L, 1);
+    }
+
+    GdkKeymapKey* keys = NULL;
+    gint n_keys;
+
```
aidanholm: no gap here

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
...
+    GdkEvent *event = gdk_event_new(GDK_KEY_PRESS);
+    GdkEventKey *event_key = (GdkEventKey *) event;
+    event_key->window = gtk_widget_get_window(w->widget);
+    event_key->send_event = TRUE;
+    event_key->time = GDK_CURRENT_TIME;
+    event_key->state = state;
+    event_key->keyval = keyval;
+    event_key->hardware_keycode = keys[0].keycode;
+    event_key->group = keys[0].group;
+    gdk_event_set_device(event, gdk_seat_get_keyboard(gdk_display_get_default_seat(gdk_display_get_default())));
```
aidanholm: super long line here; should wrap it at the least, or break it into a `GdkSeat *seat = ...` line above it.

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
...
+        MODKEY("mod3", MOD3);
+        MODKEY("mod4", MOD4);
+        MODKEY("mod5", MOD5);
+
+#undef MODKEY
+
+        lua_pop(L, 1);
+    }
+
+    GdkKeymapKey* keys = NULL;
```
aidanholm: \* goes next to the variable name: 'GdkKeymapKey *keys`

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
...
+    else
+        keyval = gdk_keyval_from_name(key_name);
+
+    if (!keyval || keyval == GDK_KEY_VoidSymbol)
+        return luaL_error(L, "failed to get a valid key value.");
+
+    guint state = 0;
+    GString *state_string = g_string_sized_new(32);
+    lua_pushnil(L);
+    while(lua_next(L, 3)) {
```
aidanholm: space between `while` and `(`

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
...
+    event_key->keyval = keyval;
+    event_key->hardware_keycode = keys[0].keycode;
+    event_key->group = keys[0].group;
+    gdk_event_set_device(event, gdk_seat_get_keyboard(gdk_display_get_default_seat(gdk_display_get_default())));
+    gdk_event_put(event);
+    debug("sending key '%s%s' to window %p", state_string->str, key_name, w);
+
+    g_string_free(state_string, TRUE);
+    g_free(keys);
+    return 1;
```
aidanholm: should be returning 0; this function doesn't return a lua value

lib/keysym.lua
```diff
@@ -0,0 +1,74 @@
...
+    local symbol = nil
+    local modifiers = {}
+    local keys = {}
+    for char in keystring:gmatch(utf8.charpattern) do
+        if char == "<" then
+            symbol = ""
+        elseif char == ">" and symbol then
+            if #symbol == 0 then
+                error("bad keystring: " .. keystring)
+                return
```
aidanholm: no need for `return` here: `error()` unwinds the stack

lib/keysym.lua
```diff
@@ -0,0 +1,74 @@
...
+-- @author Amos Bird amosbird@gmail.com
+-- @author Fabian Streitel luakit@rottenrei.be
+-- @author Mason Larobina mason.larobina@gmail.com
+-- @copyright 2017 Amos Bird amosbird@gmail.com
+-- @copyright 2010 Fabian Streitel luakit@rottenrei.be
+-- @copyright 2010 Mason Larobina mason.larobina@gmail.com
+
+local _M = {}
+
+--- Send synthetic keys to given window.
```
aidanholm, line 21: This needs more documentation of what sending keys means, and what the keystring format is. Ideally, a few examples, involving tricky cases like sending `<` and `>` keys, keys with modifiers, unicode keys..

widgets/window.c
```diff
@@ -70,6 +71,76 @@ luaH_window_set_default_size(lua_State *L)
     return 0;
 }
 
+static gint
+luaH_window_send_key(lua_State *L)
```
aidanholm, line 75: this needs some documentation added to `doc/luadoc/widget.window.lua`; can be a modified version of the keysym documentation

aidanholm (changes requested):
more minor changes needed: mostly code style and documentation

widgets/window.c
```diff
@@ -76,60 +76,52 @@ luaH_window_send_key(lua_State *L)
 {
     widget_t *w = luaH_checkwidget(L, 1);
     const gchar *key_name = luaL_checkstring(L, 2);
+    if (lua_gettop(L) >= 3) luaH_checktable(L, 3);
```
aidanholm: should have a newline after `if (...)` and `else`

widgets/window.c
```diff
@@ -76,60 +76,52 @@ luaH_window_send_key(lua_State *L)
...
-    if (!gdk_keymap_get_entries_for_keyval(gdk_keymap_get_default(), keyval, &keys, &n_keys)) {
-        lua_pushboolean(L, FALSE);
-        return 1;
-    }
-
-    g_assert(n_keys == 1);
-    g_assert(keys);
+    if (!gdk_keymap_get_entries_for_keyval(gdk_keymap_get_default(),
+                                           keyval, &keys, &n_keys))
+        return luaL_error(L, "failed to get key map entries.");
```
aidanholm: no fullstop here

doc/luadoc/widget.window.lua
```diff
@@ -15,6 +15,15 @@
 -- @tparam integer width The default width of the window, in pixels.
 -- @tparam integer height The default height of the window, in pixels.
 
+--- @method send_key
+-- Send synthetic keys to given window. This function parses a vim-like
```
aidanholm: s/synthetic keys/synthetic key events/

doc/luadoc/widget.window.lua
```diff
@@ -15,6 +15,15 @@
...
+-- keystring into single keys and sends them to the window. When
+-- `window.act_on_synthetic_keys` is on, synthetic key events will not trigger
```
aidanholm: s/on/disabled/

lib/keysym.lua
```diff
@@ -1,29 +1,48 @@
...
+-- This function parses a vim-like keystring into single keys and sends
+-- them to the window. A keystring is a string of keys to press, with
+-- special keys denoted in between angle brackets:
+--
+--     keysym.send(w, "<Shift-Home><BackSpace>")
+--     keysym.send(w, "<Control-a>")
+--     keysym.send(w, "世界，你好")
+--
+-- When `window.act_on_synthetic_keys` is on, synthetic key events will
```
aidanholm: s/on/disabled/ here as well

aidanholm (changes requested):
few more minor fixes needed

widgets/window.c
```diff
@@ -76,60 +76,55 @@ luaH_window_send_key(lua_State *L)
...
-    if (!gdk_keymap_get_entries_for_keyval(gdk_keymap_get_default(), keyval, &keys, &n_keys)) {
-        lua_pushboolean(L, FALSE);
-        return 1;
-    }
-
-    g_assert(n_keys == 1);
-    g_assert(keys);
+    if (!gdk_keymap_get_entries_for_keyval(gdk_keymap_get_default(),
+                                           keyval, &keys, &n_keys))
+        return luaL_error(L, "cannot type '%s' on current keyboard layout",
```
aidanholm: need to free `state_string` here as well, otherwise we leak memory

Issue #595, opened 2017-11-16
Follow hint letters should alternate between left- and right-hand letters


aidanholm:
Should make things faster to type. Likely implementation method: add an `s.interleave` label maker function.

PR #598, opened 2017-11-18
Interleaving url hinting.

amosbird:
Implements #595

### DIFF
--- a/lib/select_wm.lua
+++ b/lib/select_wm.lua
@@ -61,10 +61,12 @@ local label_styles = {
     end,
 
     -- Interleave style
-    interleave = function ()
+    interleave = function (left, right)
+        assert(type(left) == "string" and type(right) == "string" and #left > 0 and #right > 0, "invalid sequence")
+        assert(#left == #right, "left sequence and right sequence should be equal length")
         return function (size)
             local sub, concat = string.sub, table.concat
-            local left, right = "dsrewvcxg", "kluionmhb" -- equal length
+            local l, r, left, right = sub(left, 1, 1), sub(right, 1, 1), sub(left, 2, #left), sub(right, 2, #right)
             local base, digits, labels = #left, {}, {}
             digits[0], digits[1] = {}, {}
             for i = 1, base do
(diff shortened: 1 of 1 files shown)


lib/select_wm.lua
```diff
@@ -60,6 +60,34 @@ local label_styles = {
         return function (size) return charset("0123456789", size) end
     end,
 
+    -- Interleave style
+    interleave = function ()
+        return function (size)
+            local sub, concat = string.sub, table.concat
+            local left, right = "dsrewvcxg", "kluionmhb" -- equal length
```
aidanholm: I'd prefer that the left and right strings be passed as parameters. People might want to alternate between 1--5 and 6--0, for example. Is it possible to make this work with the parameters not equal length?

aidanholm: Also, does this function only return sequences that go LRLR...? It should also return sequences that start with a right-hand key (RLRL...)

amosbird: > Is it possible to make this work with the parameters not equal length?

It has to be equal length for now. I'll add a parameter for that with a length check then :)

> Also, does this function only return sequences that go LRLR...? It should also return sequences that start with a right-hand key (RLRL...)

it generates both LR and RL patterns

lib/select_wm.lua
```diff
@@ -60,6 +60,35 @@ local label_styles = {
...
+        return function (size)
+            local sub, concat = string.sub, table.concat
+            local base, digits, labels = #left, {}, {}
+            digits[0], digits[1] = {}, {}
+            for i = 1, base do
+                rawset(digits[1], i, sub(left, i, i))
+                rawset(digits[0], i, sub(right, i, i))
+            end
+            local maxlen = max_hint_len(ceil(size/2), base)
+            local zeroseq = string.rep("fj", maxlen)
```
aidanholm: Left-over hardcoded value here: should be `left[1] .. right[1]` I guess?

amosbird: hmm, right. I'll pop out one from left and right to use as zerochar :)

Issue #613, opened 2017-12-02
Add window.default_search_engine option


aidanholm:

McSinyx:
I've created a PR (#615) to solve this. Thanks for giving this opportunity to the community to improve our skills.

aidanholm:
Perfect! Thanks :)

PR #615, opened 2017-12-05
Add window.default_search_engine option

McSinyx:
As mentioned in issue #613.

### DIFF
--- a/lib/window.lua
+++ b/lib/window.lua
@@ -558,7 +558,7 @@ _M.methods = {
     search_open = function (_, arg)
         local lstring = lousy.util.string
         local match, find = string.match, string.find
-        local search_engines = settings.window.search_engines
+        local search_engines = settings.get_setting("window.search_engines")
 
         -- Detect blank uris
         if not arg or match(arg, "^%s*$") then return settings.window.new_tab_page end
@@ -598,9 +598,9 @@ _M.methods = {
             end
         end
 
-        -- Find search engine (or use search_engines.default)
-        local engine = "default"
-        if args[1] and args[1] ~= "default" and search_engines[args[1]] then
+        -- Find search engine (or use default_search_engine)
+        local engine = settings.get_setting("window.default_search_engine")
+        if args[1] and search_engines[args[1]] then
             engine = args[1]
             table.remove(args, 1)
         end
@@ -784,6 +784,16 @@ settings.register_settings({
             default     = "https://google.com/search?q=%s",
         },
         desc = "The set of search engine shortcuts.",
+    },
+    ["window.default_search_engine"] = {
+        type = "string",
+        default = "default",
+        validator = function (v) return settings.window.search_engines[v] end,
+        desc = [=[
+            The default search engine alias.
+
+            Must be a key of `window.search_engines`.
+        ]=],
     },
     ["window.scroll_step"] = {
         type = "number", min = 0,


lib/window.lua
```diff
@@ -559,6 +559,7 @@ _M.methods = {
         local lstring = lousy.util.string
         local match, find = string.match, string.find
         local search_engines = settings.window.search_engines
+        local engine = settings.window.default_search_engine
```
aidanholm: This should be `settings.get_setting("window.default_search_engine")` (the line above should use that function too)

lib/window.lua
```diff
@@ -599,8 +600,7 @@ _M.methods = {
         end
 
         -- Find search engine (or use search_engines.default)
-        local engine = "default"
-        if args[1] and args[1] ~= "default" and search_engines[args[1]] then
+        if args[1] and search_engines[args[1]] then
```
aidanholm: This bit of code shouldn't be changed like this, as we need to handle the `default` search engine for compatibility.

Instead: after `local engine = "default"`, but before the `if` statement, add a similar `if` statement that checks the `default_search_engine` setting for validity, and sets `engine` if it's valid

lib/window.lua
```diff
@@ -780,11 +780,19 @@ settings.register_settings({
...
         desc = "The set of search engine shortcuts.",
     },
+    ["window.default_search_engine"] = {
+        type = "string",
+        default = "google",
+        validator = function (v) return settings.window.search_engines[v] end,
+        desc = [=[
+            The default search engine alias."
+
+            Must a key of window.search_engines.
```
aidanholm: - Must be
- surround the setting name with backticks

lib/window.lua
```diff
@@ -780,11 +780,19 @@ settings.register_settings({
...
-            default     = "https://google.com/search?q=%s",
         },
         desc = "The set of search engine shortcuts.",
     },
+    ["window.default_search_engine"] = {
+        type = "string",
+        default = "google",
+        validator = function (v) return settings.window.search_engines[v] end,
+        desc = [=[
+            The default search engine alias."
```
aidanholm: Extra " at end of line

lib/window.lua
```diff
@@ -780,11 +780,19 @@ settings.register_settings({
...
             wikipedia   = "https://en.wikipedia.org/wiki/Special:Search?search=%s",
-
-            default     = "https://google.com/search?q=%s",
         },
         desc = "The set of search engine shortcuts.",
     },
+    ["window.default_search_engine"] = {
+        type = "string",
+        default = "google",
+        validator = function (v) return settings.window.search_engines[v] end,
```
aidanholm: Nice!

aidanholm (changes requested):
Thanks for the PR! Looks good: mostly just some formatting issues to fix up

lib/window.lua
McSinyx: > as we need to handle the `default` search engine for compatibility

Just to be sure that I get the idea correctly, does this mean if an user defines `search_engines` with `default` key, `default_search_engine` will be ignored?

lib/window.lua
McSinyx: Oops I was careless not double-checking the diff...

aidanholm: No worries! If you have luacheck installed, `./tests/run_test.lua luacheck` will usually catch issues like that

lib/window.lua
aidanholm: no; the other way around; the initial value is the `default` search engine, then the first `if` replaces that with the `default_search_engine` value (if valid).

aidanholm: Now that I think about it, the default value for the new setting should be `"default"`, not `"google"`; then the only change needed is to change the `local engine = ...` line to `local engine = settings.get_setting("window.default_search_engine")

McSinyx: Now I'm a bit lost:

* If the user isn't aware of the new option `default_search_engine` and keep using `search_engines["default"]` in order to achieve an equivalent result, his/her setting will be overridden. Thus this approach provides no backward-compatiblity.
* If (s)he updates the config to adapt the change, then the `default` key will only cause duplication and confusion (isn't we're trying to add the new option to avoid this awkward hard-coded table key?).

BTW should I apply the changes on separated commit or should I rebase to avoid bloated commit log?

Edit: I see you're repeating about checking the validity of `default_search_engine`'s value while  there is a validator function that can raise error when that option is set. Is checking needed anymore in this case?

aidanholm: - If they aren't aware, they won't change the new setting (it will have value `"default"`), but will have a customized `default` search engine. That works fine: the `default_search_engine` setting says "use the search engine called `default`", so the new code will work the same way.
- Updating means: changing the value of the `default_search_engine` setting, and deleting the `default` search engine entry. Normally, changes like this come with migration/deprecation warnings, which you can add if you like. There's no confusion because after this, the search engine called `default` is no longer special. It remains the default option purely by virtue of the new `default_search_engine` setting's default value.
- Same commit please :)

McSinyx: I've rebased the commits with the changes you suggested. It's a bit weird to have `default` as a search engine but escaping that (checking `args[1]`) would make it, again, special, so I hope this solution is fine.

PR #634, opened 2017-12-22
added details about keyborad event

serg-kozhemyakin:


aidanholm:
Thanks! I'd prefer that the different key properties all be collected in a single `key` table, however: `event.key.ctrl`, `event.key.alt`, `event.key.code`, etc.

serg-kozhemyakin:
i tried to mimic javascript keynames for this events. but if you think that better to have dedicated table for them -- well, why not.

aidanholm:
Thanks :)

PR #638, opened 2017-12-29
User rc file.

zappathustra:
Stop bothering about changes in rc.lua whenever you update.
I think rc.lua shouldn't be a config file at all, because who's modifying it anyway? But in the meanwhile, this could be useful.


aidanholm:
The usual reason to edit `rc.lua` is to disable some of the modules. The eventual solution is going to be a new plugin system, but this looks like a good interim approach.

BTW, you're opening a PR with your work on the `develop` branch. Normally you should make a new branch off `develop` and open a PR from there; then I can rebase that onto `develop`.

PR #639, opened 2017-12-31
show links to all luakit:// pages in the help page

jerojasro:
Hello,

I've re-installed luakit, and spent quite a good time trying to figure out how to configure it. I changed the documentation so it shows all the available chrome pages, so others don't have to dig that much to get to the config. facilities/new features available.

It's been a while since I wrote Lua; sorry for the lack of style.

Now the help page looks like this: https://imgur.com/rNt0fuV

### DIFF
--- a/lib/chrome.lua
+++ b/lib/chrome.lua
@@ -9,6 +9,7 @@
 -- @copyright 2010 Fabian Streitel <karottenreibe@gmail.com>
 
 local error_page = require("error_page")
+local lousy = require("lousy")
 local webview = require("webview")
 local window = require("window")
 local wm = require_web_module("chrome_wm")
@@ -169,6 +170,12 @@ _M.stylesheet = [===[
 local handlers = {}
 local on_first_visual_handlers = {}
 local page_funcs = {}
+
+--- Retrieve a list of the currently registered luakit:// handlers.
+-- @treturn {string} A list of `luakit://` handler names, in alphabetical order.
+function _M.available_handlers()
+    return lousy.util.table.keys(handlers)
+end
 
 --- Register a chrome page URI with an associated handler function.
 -- @tparam string page The name of the chrome page to register.
--- a/lib/help_chrome.lua
+++ b/lib/help_chrome.lua
@@ -38,11 +38,18 @@ local index_html_template = [==[
             license.  It is primarily targeted at power users, developers and any people with too much time
             on their hands who want to have fine-grained control over their web browser&rsquo;s behaviour and
             interface.</p>
-        <h2>Key bindings</h2>
+        <h2>Configuration</h2>
+        <h3>Settings</h3>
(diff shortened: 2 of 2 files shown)


lib/help_chrome.lua
```diff
@@ -78,8 +85,24 @@ local index_html_template = [==[
 </body>
 ]==]
 
+local gen_html_chrome_pages = function()
+    local links = ""
+    for _, v in pairs(chrome.available_handlers()) do
```
aidanholm: should use `ipairs` instead of `pairs`

lib/chrome.lua
```diff
@@ -170,6 +171,10 @@ local handlers = {}
 local on_first_visual_handlers = {}
 local page_funcs = {}
 
+function _M.available_handlers()
```
aidanholm: This function needs a documentation comment

aidanholm (changes requested):
Thanks, this looks great! Just some minor changes needed

jerojasro:
Thanks for checking. I've implemented the changes you suggested.

lib/chrome.lua
```diff
@@ -171,6 +171,7 @@ local handlers = {}
 local on_first_visual_handlers = {}
 local page_funcs = {}
 
+--- Returns the currently registered luakit:// handlers, as an array of strings.
```
aidanholm: - Minor quibble about phrasing, and also needs a `@treturn` line:
```
--- Retrieve a list of the currently registered luakit:// handlers.
-- @treturn {string} A list of `luakit://` handler names, in alpabetical order.
```

aidanholm (changes requested):
- Minor nit about documentation
- Please squash your fixup commits into the appropriate feature commit: the `ipairs` fixup would go into "show list of all available luakit:// pages in help page", and the doc fixup would go into the "chrome functionality one", producing two commits

Thanks!

jerojasro:
I've added the suggested changes and force-pushed them; I'll try to keep in mind that rebasing/re-writing branches for PRs is OK in this project :)

aidanholm:
Looks great, thanks!

Generally when you submit a PR (to any project), you do your work in a feature branch, and rebasing is fine because your branch is still separate from the main history, so nothing depends on it yet. In fact, rebasing is usually preferred, otherwise you end up with one or two meaningful commits and several fixups on top that clutter history. It's only once your work is merged that rebasing becomes a bad idea.

McSinyx:
I'm sorry if it's not supposed to be asked here, but why is `luakit://` chosen over `about:` in the first place?

McSinyx:
I'm sorry if it's not supposed to be asked here, but why was `luakit://` chosen over `about:` in the first place?

aidanholm:
Not a problem. That was before my time, but it may have been because webkit already uses the `about:` scheme.

PR #640, opened 2018-01-01
add forward/backward word deletion to readline.lua.

amosbird:


lib/readline.lua
```diff
@@ -49,6 +49,42 @@ local actions =  {
...
+            local i = w.ibar.input
+            local text = i.text
+            local pos = i.position
+            if text and utf8.len(text) > 1 and pos > 1 then
+                local left = string.sub(text, 2, utf8.offset(text, pos))
+                local right = string.sub(text, utf8.offset(text, pos + 1))
+                if not string.find(left, "%W+") then
+                    left = ""
+                else
+                    local _, len = string.find(left:reverse(), "%W?%w*")
```
aidanholm: Is this going to work with multibyte chars?

amosbird: Nope. Neither do all other functions in readline.lua. This is a bit tricky when combining utf8 patterns and %w. #lua irc suggests using https://github.com/starwing/luautf8

lib/readline.lua
```diff
@@ -14,6 +14,25 @@ local _M = {}
 
 local yank_ring = ""
 
+local prev_glyph = function (s, o)
+    if not o or not s or o > s:len() or o < 1 then return nil end
+    local glen = 0
+    for i = o, 1, -1 do
+        if s:sub(i, i):match("[%z\1-\x7F\xC2-\xF4]") then
```
aidanholm: This might be better as `if s:sub(i):match("^"..utf8.charpattern) then`

lib/readline.lua
```diff
@@ -14,6 +14,25 @@ local _M = {}
 
 local yank_ring = ""
 
+local prev_glyph = function (s, o)
```
aidanholm: Perhaps these functions should be in `lousy.util.string`?

PR #641, opened 2018-01-01
export history_prev/next key binding options in cmdhist.

amosbird:


aidanholm (commented):
This approach won't quite work. The values will be evaluated at module load and stored into the local `hist_binds`, so by the time you've finished loading the module it's already too late to set them!

I'd move the functions into `local history_prev_func = ...`, etc, then add a small local table `{{_M.history_prev, history_prev_func}, ... }` when the bindings are added.

lib/cmdhist.lua
```diff
@@ -16,36 +16,37 @@ local _M = {}
...
-            h.cursor = h.len
-        elseif (h.cursor or 0) > 1 then
-            h.cursor = h.cursor - 1
-        end
-        if h.cursor and h.cursor ~= lc then
-            if not h.orig then h.orig = w.ibar.input.text end
-            w:set_input(h.items[h.cursor])
-        end
-    end },
+_M.history_prev = "<Up>"
```
aidanholm: These need documentation comments, but other than that everything looks perfect

aidanholm (changes requested):

lib/cmdhist.lua
amosbird: done

PR #643, opened 2018-01-02
correct escaping of HTML entities in documentation pages

jerojasro:
when the lxsh rock is not available, the HTML escaping must be kept, or
else the text in code samples is not rendered correctly when it contains
HTML special chars.

E.g., the "FAQ" page used to look like this (html in examples does not render properly): https://imgur.com/a/J4IzQ

and now, after the change (html in code examples renders properly): https://imgur.com/a/BuBJO


aidanholm:
good find, thanks!

PR #645, opened 2018-01-04
Doc/doc bindings

jerojasro:
this pull request updates the quick start guide to remove outdated docs. about configuring luakit, and extends it a bit with examples for defining bindings.

### DIFF
--- a/doc/luadoc/pages/03-quick-start-guide.md
+++ b/doc/luadoc/pages/03-quick-start-guide.md
@@ -81,27 +81,6 @@ To customize luakit, you can define your own configuration by creating a
 is loaded automatically by luakit, if it exists. Any changes made to it will
 take effect after restarting luakit.
 
-### Location of configuration files
-
-When looking for a `.lua` file by name, luakit checks the following locations,
-in the order specified here:
-
-1. The current directory.
-2. System directories for Lua files.
-3. Luakit's collection of included modules (`/usr/share/luakit/lib/`).
-4. The user's personal luakit configuration directory (`/home/$USER/.config/luakit/`).
-5. The system configuration directories (`/etc/xdg/luakit/`).
-
-To list the directories luakit will search when loading modules, run the
-following command:
-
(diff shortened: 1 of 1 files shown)


doc/luadoc/pages/03-quick-start-guide.md
```diff
@@ -74,63 +74,69 @@ There are several different keys to scroll a webview:
...
-### Changing key bindings
+### Location of configuration files
+
+When looking for a `.lua` file by name, luakit checks the following locations,
+in the order specified here:
+
+1. The current directory.
+2. System directories for Lua files.
+3. Luakit's collection of included modules (`/usr/share/luakit/lib/`).
+4. The user's personal luakit configuration directory (`/home/$USER/.config/luakit/`).
```
McSinyx: I think [this should be `$XDG_CONFIG_DIR/luakit/`](https://github.com/luakit/luakit/blob/1ceb5fdc81af135516725dd61544e1e739ba4230/globalconf.h#L39).

jerojasro: (initial disclaimer: I did not intend to modify that text, only the text related to bindings; it shows up in the diff because I was reorganizing paragraphs.)

I'm not so sure; I know that's the way luakit looks at it, but it's not the way a regular user thinks of the location of that file. In my machine (debian) that environment variable is not defined, in my user account.

I think it's best to talk to the user here, and maybe mention that the way the files are looked up follows the XDG config. file schema, but still tell the user where to put the file in a way that only requires basic familiarity with the command line, instead of knowledge about XDG.

aidanholm: - The usual terminology is "config directory", as defined in `luakit://help/doc/pages/05-configuration.html`, with no example configuration path.
- If you are going to use example paths, which is probably fine as it's a quickstart guide, I'd prefer `~/foo/` to `/home/$USER/foo`.

aidanholm:
Thanks, looks good!

PR #646, opened 2018-01-05
Improve mechanism of opening selection contents

### LINKED ISSUE #642: Should content of clipboard/primary be splited to words when opened using p* or P*?
I want to discuss the logic behind these bindings. I'm not sure about other's use of this binding, but I mainly use it when I need to search for terms in articles I read. IMHO, it's very unlikely that one would ever want to search for each selected word, while difficult terms are usually phrases. Let's take an example from [debian.org](https://www.debian.org/license) to make it clearer:

> Work is in progress to make the older material compliant with the above licenses. Until then, please refer to the following terms of the Open Publication License.
>
> This material may be distributed only subject to the terms and conditions set forth in the Open Publication License, Draft v1.0 or later (you can read our local copy, the latest version is usually available at http://www.opencontent.org/openpub/).
>
> “Debian” and the Debian Logo are trademarks of Software in the Public Interest, Inc.

I can't speak for everyone, but `pp` in this case would make a lot more sense if it can be used to search for group of words such as *Open Publication License* or *Public Interest, Inc.*. I can imagine [the current implementation](https://github.com/luakit/luakit/blob/6bb3df7f57f8a8131d522998a42f7d6f25af7fb3/lib/binds.lua#L277) was made (7991929ac66c42516b0b8c14bf7a283567704436) for the following kind of case, but this would be a rare case compare to the above example:

> http://www.jclark.com/xml/copying.txt
> http://www.gnu.org/licenses/gpl.html

Furthermore, it's quite irritating to accidentally press this key combination if one have the habit of selecting while reading - then dozens of tab may be opened.

McSinyx:
Fixes #642.

Although `lousy.uri.split` is only used in `binds`, I keep it in `lousy.uri` to allow users to use it in their config (e.g. `:open` with `pp` mechanism).

### DIFF
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -241,16 +241,13 @@ modes.add_binds("normal", {
     -- Open primary selection contents.
     { "pp", [[Open URLs based on the current primary selection contents in the current tab.]],
         function (w)
-            local uris = {}
-            for uri in string.gmatch(luakit.selection.primary or "", "%S+") do
-                table.insert(uris, uri)
-            end
-            if #uris == 0 then w:notify("Nothing in primary selection...") return end
-            w:navigate(w:search_open(uris[1]))
-            if #uris > 1 then
-                for i=2,#uris do
-                    w:new_tab(w:search_open(uris[i]))
-                end
+            local uris = lousy.uri.split(luakit.selection.primary or "")
+            i, uri = next(uris)
+            if not i then w:notify("Nothing in primary selection...") return end
+            w:navigate(w:search_open(uri))
+            while next(uris, i) do
+                i, uri = next(uris, i)
+                w:new_tab(w:search_open(uri))
             end
         end },
     { "pt", [[Open a URL based on the current primary selection contents in `[count=1]` new tab(s).]],
@@ -261,32 +258,26 @@ modes.add_binds("normal", {
         end, {count = 1} },
     { "^pw$", [[Open URLs based on the current primary selection contents in a new window.]],
         function(w)
-            local uris = {}
-            for uri in string.gmatch(luakit.selection.primary or "", "%S+") do
-                table.insert(uris, uri)
-            end
-            if #uris == 0 then w:notify("Nothing in primary selection...") return end
-            w = window.new{w:search_open(uris[1])}
-            if #uris > 1 then
-                for i=2,#uris do
-                    w:new_tab(w:search_open(uris[i]))
-                end
+            local uris = lousy.uri.split(luakit.selection.primary or "")
+            i, uri = next(uris)
+            if not i then w:notify("Nothing in primary selection...") return end
+            w = window.new{w:search_open(uri)}
+            while next(uris, i) do
+                i, uri = next(uris, i)
+                w:new_tab(w:search_open(uri))
             end
         end },
 
     -- Open clipboard contents.
     { "^PP$", [[Open URLs based on the current clipboard selection contents in the current tab.]],
         function (w)
-            local uris = {}
-            for uri in string.gmatch(luakit.selection.clipboard or "", "%S+") do
-                table.insert(uris, uri)
-            end
-            if #uris == 0 then w:notify("Nothing in clipboard...") return end
-            w:navigate(w:search_open(uris[1]))
-            if #uris > 1 then
-                for _=2,#uris do
-                    w:new_tab(w:search_open(uris[1]))
-                end
+            local uris = lousy.uri.split(luakit.selection.clipboard or "")
+            i, uri = next(uris)
+            if not i then w:notify("Nothing in clipboard...") return end
+            w:navigate(w:search_open(uri))
+            while next(uris, i) do
+                i, uri = next(uris, i)
+                w:new_tab(w:search_open(uri))
             end
         end },
 
@@ -299,16 +290,13 @@ modes.add_binds("normal", {
 
     { "^PW$", [[Open URLs based on the current clipboard selection contents in a new window.]],
         function(w)
-            local uris = {}
-            for uri in string.gmatch(luakit.selection.clipboard or "", "%S+") do
-                table.insert(uris, uri)
-            end
-            if #uris == 0 then w:notify("Nothing in clipboard...") return end
-            w = window.new{w:search_open(uris[1])}
-            if #uris > 1 then
-                for i=2,#uris do
-                    w:new_tab(w:search_open(uris[i]))
-                end
+            local uris = lousy.uri.split(luakit.selection.clipboard or "")
+            i, uri = next(uris)
+            if not i then w:notify("Nothing in clipboard...") return end
+            w = window.new{w:search_open(uri)}
+            while next(uris, i) do
+                i, uri = next(uris, i)
+                w:new_tab(w:search_open(uri))
             end
         end },
 
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -247,8 +247,8 @@ modes.add_binds("normal", {
             local engine = settings.get_setting("window.default_search_engine")
             local uris = split_uri(luakit.selection.primary or "")
             if #uris == 0 then w:notify("Nothing in primary selection...") return end
-            local uri = table.remove(uris, 1)
-            w:navigate(is_uri(uri) and uri or w:search_open(engine .. uri))
+            local uri1 = table.remove(uris, 1)
+            w:navigate(is_uri(uri1) and uri1 or w:search_open(engine .. uri1))
             for _, uri in ipairs(uris) do
                 w:new_tab(is_uri(uri) and uri or w:search_open(engine .. uri))
             end
@@ -267,8 +267,8 @@ modes.add_binds("normal", {
             local engine = settings.get_setting("window.default_search_engine")
             local uris = split_uri(luakit.selection.primary or "")
             if #uris == 0 then w:notify("Nothing in primary selection...") return end
-            local uri = table.remove(uris, 1)
-            w = window.new{is_uri(uri) and uri or w:search_open(engine .. uri)}
+            local uri1 = table.remove(uris, 1)
+            w = window.new{is_uri(uri1) and uri1 or w:search_open(engine .. uri1)}
             for _, uri in ipairs(uris) do
                 w:new_tab(is_uri(uri) and uri or w:search_open(engine .. uri))
             end
@@ -280,8 +280,8 @@ modes.add_binds("normal", {
             local engine = settings.get_setting("window.default_search_engine")
             local uris = split_uri(luakit.selection.clipboard or "")
             if #uris == 0 then w:notify("Nothing in primary selection...") return end
-            local uri = table.remove(uris, 1)
-            w:navigate(is_uri(uri) and uri or w:search_open(engine .. uri))
+            local uri1 = table.remove(uris, 1)
+            w:navigate(is_uri(uri1) and uri1 or w:search_open(engine .. uri1))
             for _, uri in ipairs(uris) do
                 w:new_tab(is_uri(uri) and uri or w:search_open(engine .. uri))
             end
@@ -300,8 +300,8 @@ modes.add_binds("normal", {
             local engine = settings.get_setting("window.default_search_engine")
             local uris = split_uri(luakit.selection.clipboard or "")
             if #uris == 0 then w:notify("Nothing in clipboard...") return end
-            local uri = table.remove(uris, 1)
-            w = window.new{is_uri(uri) and uri or w:search_open(engine .. uri)}
+            local uri1 = table.remove(uris, 1)
+            w = window.new{is_uri(uri1) and uri1 or w:search_open(engine .. uri1)}
             for _, uri in ipairs(uris) do
                 w:new_tab(is_uri(uri) and uri or w:search_open(engine .. uri))
             end
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -22,6 +22,9 @@ local modes = require("modes")
 -- Util aliases
 local join, split = lousy.util.table.join, lousy.util.string.split
 
+-- URI aliases
+local is_uri, split_uri = lousy.uri.is_uri, lousy.uri.split
+
 --- Compatibility wrapper for @ref{modes/add_binds|modes.add_binds()}.
 -- @deprecated use @ref{modes/add_binds|modes.add_binds()} instead.
 _M.add_binds = function (...)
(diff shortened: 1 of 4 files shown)


McSinyx:
Wait a minute, should `pt` and `PT` have the same behavior as well?

Edit: I added a commit to uniform their bahaviors.

McSinyx:
@aidanholm Could you please take a look at this PR :bowing_man:? I've been quite nervous for the last few days as I'm still very new in real-life software development.

lib/binds.lua
```diff
@@ -239,76 +239,70 @@ modes.add_binds("normal", {
     { "<F11>", "Toggle fullscreen mode.", function (w) w.win.fullscreen = not w.win.fullscreen end },
 
     -- Open primary selection contents.
-    { "pp", [[Open URLs based on the current primary selection contents in the current tab.]],
+    { "^pp$", [[Open URLs based on the current primary selection contents in the current tab.]],
```
aidanholm: There should be no need to add ^ and $ to the pattern. Same for all the other patterns.

lib/lousy/uri.lua
```diff
@@ -56,6 +56,55 @@ local opts_metatable = {
     end,
 }
 
+--- Guess if a string is an URI.
+-- @tparam string s The input string.
+-- @treturn boolean true if the string could be a URI, false otherwise.
+function _M.is_uri(s)
+    if s == "about:blank" then return true end
+
+    -- Navigate if the string contains . or /; or if it is a javascript-uri
```
aidanholm: 1) Don't need the comment; it's obvious what the code does
2) I realise the old code was like this, but is . or /  a good heuristic? I can think of search terms that might include a . Maybe use `lousy.uri.parse`? (That doesn't work for about:blank or javascript: uris IIRC, which is why we need a separate `is_uri` function)

lib/window.lua
```diff
@@ -581,11 +581,10 @@ _M.methods = {
     -- Intelligent open command which can detect a uri or search argument.
     search_open = function (_, arg)
         local lstring = lousy.util.string
-        local match, find = string.match, string.find
         local search_engines = settings.get_setting("window.search_engines")
 
         -- Detect blank uris
-        if not arg or match(arg, "^%s*$") then return settings.get_setting("window.new_tab_page") end
+        if not arg or arg:match("^%s*$") then return settings.get_setting("window.new_tab_page") end
```
aidanholm, line 587: This change is not relevant, so it shouldn't be included in the commit.

lib/lousy/uri.lua
```diff
@@ -56,6 +56,55 @@ local opts_metatable = {
...
+
+    return false
+end
+
+--- Split a string into URIs or group of words.
+-- @tparam string s The string to split.
+-- @treturn table A table of strings whose member is either a group of words to
+-- be searched or a URI
+function _M.split(s)
+    local get_setting = require("settings").get_setting
```
aidanholm: I don't think that this function should add the search engine at all. Inserting the search engine at the beginning of search terms is specific to opening a bunch of split URIs/search terms, but is _not_ specific to the splitting itself.

I.e. search engines should be handled by the bindings.

lib/binds.lua
```diff
@@ -239,76 +239,70 @@ modes.add_binds("normal", {
...
-                table.insert(uris, uri)
-            end
-            if #uris == 0 then w:notify("Nothing in primary selection...") return end
-            w:navigate(w:search_open(uris[1]))
-            if #uris > 1 then
-                for i=2,#uris do
-                    w:new_tab(w:search_open(uris[i]))
-                end
+            local uris = lousy.uri.split(luakit.selection.primary or "")
+            i, uri = next(uris)
```
aidanholm: This needs `local` in front, as do all the other bits of similar code in the other bindings.

aidanholm (changes requested):
Hi @McSinyx, sorry for the delay! This looks mostly good, but there are a few changes needed.

`tests/async/test_lib_lousy_uri.lua` is also failing and needs to be edited, but that should be easy.

Also, it'd be great if you could write some small tests for `lousy.uri.is_uri` and `lousy.uri.split` (also in tests/async/test_lib_lousy_uri.lua)

lib/binds.lua
McSinyx: Thank you, I tried to uniform the format from `*` to `^*$` while it was supposed to be the opposite way.

lib/binds.lua
McSinyx: Oops, I haven't got used to Lua strict scopes yet.

lib/lousy/uri.lua
McSinyx: 2. Most search engines doesn't support symbol, except Symbol Hound. Well, that is the point: we always try to be too smart, but the function is too guess any way, and search can  always be enforced by user using `:open google search-string`. In addition, the problem in the logic here is that we need to grep URI outta articles, so URIs are not always  surrounded be whitespaces (i.e. followed by comma, period or in side brackets). Maybe I'll add *pseudo* to the function name, but it's already in `lousy` anyway.

lib/lousy/uri.lua
McSinyx: That's a good point, although `is_url` will need to be called again in bindings (which IMHO doesn't affect performance much).

lib/window.lua
McSinyx: This is because I removed [L584](https://github.com/luakit/luakit/pull/646/files/777ff664390e2c0321fd1ff7d188750da82421ca#diff-ee6e38c430e943fe52665d5cc6002dafL584), hope it doesn't create too much noise.

McSinyx:
Thanks, I'll do the edits and create tests when I get home this afternoon.

lib/binds.lua
aidanholm: Yeah, global-by-default is a bit of a pain, since you almost always have to use `local`, but it's necessary due to lua's lexical scoping.

lib/lousy/uri.lua
aidanholm: Should be fine: it would only an issue if someone had lots of hosts

lib/lousy/uri.lua
aidanholm: Stripping junk like parens or commas from URIs belongs in `_M.split()`, although it'll make that function more complicated: probably something like this:
```lua
local junk = "([%.%{%}%,%[%]%<%>%;%:]*)"
local _, uri, _ = p:match("^"..junk..".-"..junk.."$")
if uri and _M.is_uri(uri) then ... elseif _M.is_uri(p) then ... else ... end
```

lib/window.lua
aidanholm: It'd be better to change L584 to only remove `find` then. In this case, it's not terribly important, but the principle is that when you're trying to understand a line of code by using `git blame`, irrelevant changes make that take longer.

McSinyx:
I'm so sorry to say that it's now my turn to apologize for the delay, as lately I've been quite busy on my undergrad application (tbh I'm worried af, and it feels good to put all those a side and do something else less stressful like this).

There are a few things that I'm still having trouble on:

1. Somehow `make` doesn't compile `tests/uri.c` (?)
2. IMHO `lousy.util.is_uri` is alright now with these possible cases:
    i. Full URI, e.g. `https://github.com`
    ii. W/o scheme, e.g. `debian.org`
    iii. File paths, e.g. `/home`. `Downloads/` or `.config`
    iv. Local host, e.g. `localhost:8000`
3. Brackets can be striped if they matches (I mean something similiar to `[(example.com)]`), but period and commas are, well, more problematic. Somehow trailing period is (semi-)legal ([http://www.shareprice.co.uk/TW.](http://www.shareprice.co.uk/TW.)) and comma can exist within the option part or the URL, so it can end the link as well.

Edit: about 2.iii, we should be stricter on filename. Personally I think trailing slash should be required, e.g. `.config/` is OK while `.config` should be searched.

lib/binds.lua
```diff
@@ -241,74 +244,74 @@ modes.add_binds("normal", {
...
-            w = window.new{w:search_open(uris[1])}
-            if #uris > 1 then
-                for i=2,#uris do
-                    w:new_tab(w:search_open(uris[i]))
-                end
+    { "pt", [[Open URLs based on the current primary selection contents in new tabs.]],
+        function (w)
+            local engine = settings.get_setting("window.default_search_engine")
+            local uris = split_uri(luakit.selection.primary or "")
+            local i, uri = next(uris)
```
aidanholm: split_uri returns an array (i.e. it has numeric keys from 1..n), so why not just check the length, and use a loop?
```lua
if #uris == 0 then w:notify("...") return end
for _, uri in ipairs(uris) do ... end
```

lib/binds.lua
```diff
@@ -241,74 +244,74 @@ modes.add_binds("normal", {
...
+                i, uri = next(uris, i)
+            until i == nil
+        end },
+    { "pw", [[Open URLs based on the current primary selection contents in a new window.]],
+        function (w)
+            local engine = settings.get_setting("window.default_search_engine")
+            local uris = split_uri(luakit.selection.primary or "")
+            local i, uri = next(uris)
+            if not i then w:notify("Nothing in primary selection...") return end
+            w = window.new{is_uri(uri) and uri or w:search_open(engine .. uri)}
```
aidanholm: Same here: you can use `table.remove` to pull out the first URI to open the window with, and then a for loop over the rest

lib/binds.lua
```diff
@@ -241,74 +244,74 @@ modes.add_binds("normal", {
...
-            end
-            if #uris == 0 then w:notify("Nothing in clipboard...") return end
-            w:navigate(w:search_open(uris[1]))
-            if #uris > 1 then
-                for _=2,#uris do
-                    w:new_tab(w:search_open(uris[1]))
-                end
+            local engine = settings.get_setting("window.default_search_engine")
+            local uris = split_uri(luakit.selection.clipboard or "")
+            local i, uri = next(uris)
```
aidanholm: ditto

lib/binds.lua
```diff
@@ -241,74 +244,74 @@ modes.add_binds("normal", {
...
-            w = window.new{w:search_open(uris[1])}
-            if #uris > 1 then
-                for i=2,#uris do
-                    w:new_tab(w:search_open(uris[i]))
-                end
+    { "PT", [[Open URLs based on the current clipboard selection contents in new tabs.]],
+        function (w)
+            local engine = settings.get_setting("window.default_search_engine")
+            local uris = split_uri(luakit.selection.clipboard or "")
+            local i, uri = next(uris)
```
aidanholm: ditto

lib/binds.lua
```diff
@@ -241,74 +244,74 @@ modes.add_binds("normal", {
...
+            repeat
+                w:new_tab(is_uri(uri) and uri or w:search_open(engine .. uri))
+                i, uri = next(uris, i)
+            until i == nil
+        end },
+    { "PW", [[Open URLs based on the current clipboard selection contents in a new window.]],
+        function (w)
+            local engine = settings.get_setting("window.default_search_engine")
+            local uris = split_uri(luakit.selection.clipboard or "")
+            local i, uri = next(uris)
```
aidanholm: ditto

aidanholm (changes requested):

aidanholm:
No worries; we all have limited time; good luck on your undergrad applications!

If there are compile errors, please post any error messages. I don't really have an opinion on how picky `is_uri` is: it's never going to be right 100% of the time. I do think that periods and commas should be stripped, as in my experience they're usually superfluous; and this only strips from the ends, not the middle, so it's easy to manually fix for the rare case where it's needed.

McSinyx:
Well, after running `make`, there still isn't `tests/util.so` so `run_test.lua` fail:

```sh
$ ./tests/run_test.lua luacheck
luajit: ./tests/run_test.lua:15: module 'tests.util' not found:
        no field package.preload['tests.util']
        no file './tests/util.lua'
        no file '/usr/share/luajit-2.1.0-beta3/tests/util.lua'
        no file '/usr/local/share/lua/5.1/tests/util.lua'
        no file '/usr/local/share/lua/5.1/tests/util/init.lua'
        no file '/usr/share/lua/5.1/tests/util.lua'
        no file '/usr/share/lua/5.1/tests/util/init.lua'
        no file './tests/tests/util.lua'
        no file './lib/tests/util.lua'
        no file './lib/tests/util/init.lua'
        no file './tests/util.so'
        no file '/usr/local/lib/lua/5.1/tests/util.so'
        no file '/usr/lib/i386-linux-gnu/lua/5.1/tests/util.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
        no file './tests.so'
        no file '/usr/local/lib/lua/5.1/tests.so'
        no file '/usr/lib/i386-linux-gnu/lua/5.1/tests.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
        [C]: in function 'require'
        ./tests/run_test.lua:15: in main chunk
        [C]: at 0x0048b110
$ ls tests/
async  html  lib.lua  priv.lua  run_test.lua  style  util.c
```

aidanholm:
do you have `DEVELOPMENT_PATHS` set to 1 in your config.mk?

McSinyx:
:stuck_out_tongue: silly me, I didn't. But wait, there is still not `tests/util.so`. I guess I'll use the generated CFLAGS from `make` to compile it.

<details>
<summary><strike>Edit: that doesn't work either</strike></summary>
$ gcc -std=gnu99 -ggdb -W -Wall -Wextra -Werror=unused-result -pthread -I/
usr/include/webkitgtk-4.0 -I/usr/include/gtk-3.0 -I/usr/include/at-spi2-atk/2.0 -I/usr/include/at-s
pi-2.0 -I/usr/include/dbus-1.0 -I/usr/lib/i386-linux-gnu/dbus-1.0/include -I/usr/include/gtk-3.0 -I
/usr/include/gio-unix-2.0/ -I/usr/include/cairo -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I
/usr/include/pango-1.0 -I/usr/include/atk-1.0 -I/usr/include/cairo -I/usr/include/pixman-1 -I/usr/i
nclude/freetype2 -I/usr/include/libpng16 -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/libpng16 -I/u
sr/include/libsoup-2.4 -I/usr/include/libxml2 -I/usr/include/luajit-2.1 -I/usr/include/webkitgtk-4.
0 -I/usr/include/glib-2.0 -I/usr/lib/i386-linux-gnu/glib-2.0/include -I./ -o tests/util.so tests/ut
il.c
/usr/lib/gcc/i686-linux-gnu/7/../../../i386-linux-gnu/Scrt1.o: In function `_start':
(.text+0x28): undefined reference to `main'
/tmp/ccv42Qta.o: In function `l_getenv':
/home/cnx/Sources/luakit/tests/util.c:83: undefined reference to `luaL_checklstring'
/home/cnx/Sources/luakit/tests/util.c:83: undefined reference to `g_getenv'
/home/cnx/Sources/luakit/tests/util.c:83: undefined reference to `lua_pushstring'
/tmp/ccv42Qta.o: In function `l_spawn_async':
/home/cnx/Sources/luakit/tests/util.c:46: undefined reference to `lua_type'
/home/cnx/Sources/luakit/tests/util.c:46: undefined reference to `luaL_typerror'
/home/cnx/Sources/luakit/tests/util.c:47: undefined reference to `lua_objlen'
/home/cnx/Sources/luakit/tests/util.c:50: undefined reference to `g_ptr_array_sized_new'
/home/cnx/Sources/luakit/tests/util.c:57: undefined reference to `lua_tolstring'
/home/cnx/Sources/luakit/tests/util.c:57: undefined reference to `g_ptr_array_add'
/home/cnx/Sources/luakit/tests/util.c:58: undefined reference to `lua_settop'
/home/cnx/Sources/luakit/tests/util.c:52: undefined reference to `lua_rawgeti'
/home/cnx/Sources/luakit/tests/util.c:53: undefined reference to `lua_isstring'
/home/cnx/Sources/luakit/tests/util.c:54: undefined reference to `g_ptr_array_free'
/home/cnx/Sources/luakit/tests/util.c:55: undefined reference to `luaL_error'
/home/cnx/Sources/luakit/tests/util.c:49: undefined reference to `luaL_error'
/home/cnx/Sources/luakit/tests/util.c:60: undefined reference to `g_ptr_array_add'
/home/cnx/Sources/luakit/tests/util.c:66: undefined reference to `g_spawn_async'
/home/cnx/Sources/luakit/tests/util.c:68: undefined reference to `g_ptr_array_free'
/home/cnx/Sources/luakit/tests/util.c:73: undefined reference to `lua_pushnil'
/home/cnx/Sources/luakit/tests/util.c:74: undefined reference to `lua_pushstring'
/home/cnx/Sources/luakit/tests/util.c:75: undefined reference to `g_error_free'
/home/cnx/Sources/luakit/tests/util.c:71: undefined reference to `lua_pushnumber'
/tmp/ccv42Qta.o: In function `l_make_tmp_dir':
/home/cnx/Sources/luakit/tests/util.c:30: undefined reference to `lua_type'
/home/cnx/Sources/luakit/tests/util.c:32: undefined reference to `g_dir_make_tmp'
/home/cnx/Sources/luakit/tests/util.c:33: undefined reference to `lua_pushstring'
/home/cnx/Sources/luakit/tests/util.c:34: undefined reference to `g_free'
/home/cnx/Sources/luakit/tests/util.c:36: undefined reference to `lua_pushstring'
/home/cnx/Sources/luakit/tests/util.c:37: undefined reference to `g_error_free'
/home/cnx/Sources/luakit/tests/util.c:30: undefined reference to `luaL_checklstring'
/tmp/ccv42Qta.o: In function `l_kill':
/home/cnx/Sources/luakit/tests/util.c:90: undefined reference to `luaL_checknumber'
/home/cnx/Sources/luakit/tests/util.c:91: undefined reference to `luaL_optinteger'
/home/cnx/Sources/luakit/tests/util.c:95: undefined reference to `lua_pushstring'
/tmp/ccv42Qta.o: In function `luaopen_tests_util':
/home/cnx/Sources/luakit/tests/util.c:109: undefined reference to `luaL_openlib'
collect2: error: ld returned 1 exit status
</details><br>

Edit 2: I've just read the `Makefile`, there is `tests/util.so` target. Everything works now, I'm sorry for such a silly question.

McSinyx:
I'm so sorry that I messed up the tree while rebasing :confused:

I've created PR #654 to replace this.

McSinyx:
I'm back here just to say thank you @aidanholm for your blessing: I got accepted after the interview this morning.

aidanholm:
Congratulations!

McSinyx:
Thanks again man! It feels so relieving to be able to put away the worries.

PR #654, opened 2018-01-14
Improve mechanism of opening selection contents

### LINKED ISSUE #642: Should content of clipboard/primary be splited to words when opened using p* or P*?

McSinyx:
Fix #642 (sorry I messed up the local `fix-642` branch).

### DIFF
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -22,6 +22,9 @@ local modes = require("modes")
 -- Util aliases
 local join, split = lousy.util.table.join, lousy.util.string.split
 
+-- URI aliases
+local is_uri, split_uri = lousy.uri.is_uri, lousy.uri.split
+
 --- Compatibility wrapper for @ref{modes/add_binds|modes.add_binds()}.
 -- @deprecated use @ref{modes/add_binds|modes.add_binds()} instead.
 _M.add_binds = function (...)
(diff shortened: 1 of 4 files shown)


aidanholm:
Thanks! And don't worry about it :)

PR #656, opened 2018-01-14
Add settings for completion

zappathustra:
Create settings to control completion:

- `completion.history_order` to sort items according to the database columns: visits, last_visit, title, uri (not sure the last two are interesting, but they don't hurt).
- `completion.max_items` to limit the number of items displayed.

More could be done, but that's what I needed at least.

### DIFF
--- a/lib/completion.lua
+++ b/lib/completion.lua
@@ -314,8 +314,6 @@ settings.register_settings({
     }
 })
 
-
-
 completers.history = {
     header = { "History", "URI" },
     func = function (buf)
--- a/lib/completion.lua
+++ b/lib/completion.lua
@@ -290,14 +290,42 @@ local function sql_like_globber(term)
     return "%" .. escaped:gsub("%s+", "%%") .. "%"
 end
 
+settings.register_settings({
+    ["completion.history_order"] = {
+        type = "string",
+        default = "visits",
+        desc = [=[
+            A string indicating how history items should be sorted in
+            completion. Possible values are:<br>
+            - visits:     most visited URI's first<br>
+            - last_visit: most recent URI's first<br>
+            - title:      alphabetical sort<br>
+            - uri:        alphabetical sort
+        ]=],
+        validator = function (v)
+            local t = {visits = true, last_visit = true, title = true, uri = true}
+            return t[v]
+        end
+    },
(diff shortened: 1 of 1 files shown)


lib/completion.lua
```diff
@@ -290,14 +290,42 @@ local function sql_like_globber(term)
...
+    },
+    ["completion.max_items"] = {
+        type = "number", min = 1,
+        default = 25,
+        desc = "Number of completion items for history and bookmarks."
+    }
+})
+
+
+
```
McSinyx: Man, why 2 trailing newlines:question:

zappathustra: Bad me :)
Cleaned up.

lib/completion.lua
```diff
@@ -290,14 +290,40 @@ local function sql_like_globber(term)
     return "%" .. escaped:gsub("%s+", "%%") .. "%"
 end
 
+settings.register_settings({
+    ["completion.history_order"] = {
```
aidanholm: I'd prefer `completion.history.order` here

lib/completion.lua
```diff
@@ -290,14 +290,40 @@ local function sql_like_globber(term)
...
+        type = "string",
+        default = "visits",
+        desc = [=[
+            A string indicating how history items should be sorted in
+            completion. Possible values are:<br>
```
aidanholm: - desc uses markdown, so you shouldn't need to use `<br>` tags
- the keyswords (`visits`, `last_visit`, etc.) should be wrapped in backticks so they have code formatting

lib/completion.lua
```diff
@@ -290,14 +290,40 @@ local function sql_like_globber(term)
...
+            - visits:     most visited URI's first<br>
+            - last_visit: most recent URI's first<br>
+            - title:      alphabetical sort<br>
+            - uri:        alphabetical sort
+        ]=],
+        validator = function (v)
+            local t = {visits = true, last_visit = true, title = true, uri = true}
```
aidanholm: :+1:

aidanholm (commented):
This looks pretty great! Just some formatting nits

aidanholm:
Thanks! Merged this one manually

Issue #620, opened 2017-12-06
Add warning for deprecated files in config dir


aidanholm:
E.g. binds, modes, others...
Shouldn't warn if config directory has higher priority than system lib dir in `package.path`.

See #619

PR #657, opened 2018-01-14
Warning for deprecated config files (see #620).

zappathustra:
This addresses #620.

Overloading `require()` might not be a good idea, but this is a first draft. Also, I guess the warning could be reworded.

### DIFF
--- a/config/rc.lua
+++ b/config/rc.lua
@@ -4,23 +4,19 @@
 
 require "lfs"
 
--- Checks whether there exist module files in the local config dir and warns
--- that it's deprecated, unless you really want to use them. It's not really a
--- searchers function, it simply issues a warning, then the next function in
--- package.searchers is used to load the module.
-local searchers = package.searchers or package.loaders
-local function deprecated_files (module)
-    local f = package.searchpath(module, package.path)
-    local lf = luakit.config_dir .. "/" .. module .. ".lua"
-    if f ==  lf then
-        msg.warn("Loading local version of '" .. module .. "' module: " .. lf)
+-- Check for lua configuration files that will never be loaded because they are
+-- shadowed by builtin modules.
+table.insert(package.loaders, 2, function (modname)
+    local f = package.searchpath(modname, package.path)
+    local lf = luakit.config_dir .. "/" .. modname .. ".lua"
+    if f == lf then
+        msg.warn("Loading local version of '" .. modname .. "' module: " .. lf)
     elseif lfs.attributes(lf) then
         msg.warn("Found local version " .. lf
-            .. " for core module '" .. module
+            .. " for core module '" .. modname
             .. "', but it won't be used, unless you update 'package.path' accordingly.")
     end
-end
-table.insert(searchers, 2, deprecated_files) 
+end)
 
 require "unique_instance"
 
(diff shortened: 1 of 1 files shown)


aidanholm:
Hm, I think I agree with you that it's not a good idea :) A better solution is to insert a function at the beginning of `package.loaders`, which is what `require` calls to load modules.

zappathustra:
Oh, yes, `package.loaders` is a much more sensible choice! I'll implement that. For the record, it's `package.searchers` since 5.2, so I'll do something like this:

```
local loaders = package.searchers or package.loaders
```

so there'll be no problem when (if) you move to more recent Lua versions.

zappathustra:
Note: this will warn if there exists a non-luakit lua module (say LFS) in the config dir. To make sure that we're only checking for the files in `$INSTALL_DIR/lib`, a more complicated scheme is required, but is it really useful?

I noticed also that `luakit.install_paths.(install|config)_dir` do not return `./(lib|config)` when `$DEVELOPMENT_PATHS` is 1. Is that normal? That makes those values rather unusable when running from source.

config/rc.lua
```diff
@@ -161,6 +179,16 @@ local tab_favicons = require "tab_favicons"
 -- Add :view-source command
 local view_source = require "view_source"
 
+-- Perhaps paranoid, but you never know.
```
aidanholm: I think this is probably a little too paranoid; ordinarily I'd appreciate such caution, but here luakit controls the initial searchers (we can modify them before loading `rc.lua` if the need arises), so they effectively form part of the luakit API. The other reason is that this is going in the default rc file, which should be kept as short as possible.

config/rc.lua
```diff
@@ -4,6 +4,24 @@
 
 require "lfs"
 
+-- Checks whether there exist module files in the local config dir and warns
+-- that it's deprecated, unless you really want to use them. It's not really a
+-- searchers function, it simply issues a warning, then the next function in
+-- package.searchers is used to load the module.
+local searchers = package.searchers or package.loaders
+local function deprecated_files (module)
```
aidanholm: for the sake of brevity, I'd actually prefer to see this made an anonymous function, like this:
```lua
table.insert(package.loaders, function (modname)
    ...
end)
```
I think just using `package.loaders` is fine; if compatibility ever becomes an issue, we can deal with it then.
Also, `module` is already a global (although it might be good to clear it before `rc.lua` runs...) and it's better not to shadow it; `modname` is the name the lua51 api docs use.

config/rc.lua
```diff
@@ -4,6 +4,24 @@
 
 require "lfs"
 
+-- Checks whether there exist module files in the local config dir and warns
```
aidanholm: I appreciate the extensive documentation but think this could be shortened a bit, as it's part of the `rc.lua` file:
"Check for lua configuration files that will never be loaded because that are shadowed by builtin modues."

config/rc.lua
```diff
@@ -4,6 +4,24 @@
...
+-- that it's deprecated, unless you really want to use them. It's not really a
+-- searchers function, it simply issues a warning, then the next function in
+-- package.searchers is used to load the module.
+local searchers = package.searchers or package.loaders
+local function deprecated_files (module)
+    local f = package.searchpath(module, package.path)
+    local lf = luakit.config_dir .. "/" .. module .. ".lua"
+    if f ==  lf then
```
aidanholm: spacing

aidanholm (changes requested):
Sorry for the long delay! :(

config/rc.lua
zappathustra: What I found paranoid is the way I remove the function, not the removal itself (necessary otherwise `userconf.lua`, and files required in it, produce warnings). Is that what you mean too, so you'd rather do `table.remove(package.loaders, 2)` instead?

aidanholm:
Thanks! merged :)

zappathustra:
You did not include the `table.remove` line in your merge. So there is a warning when `userconf.lua` is loaded. Is that on purpose?

aidanholm:
That was an oversight on my part, but I've realized that the current implementation is not ideal. It warns for all local modules, but should only warn for modules that shadow a builtin module. I'll fix it up a bit in develop.

aidanholm:
Added at 817586d

PR #659, opened 2018-01-15
Add permanent history to command-line

zappathustra:
This makes command/lua modes history permanent, i.e. it persists between invocations. It requires creating the `luakit.data_dir/(command|lua)-history` files.

It would be more elegant to rewrite the files when Luakit closes, but I wanted to keep the modifications where history operations already appear, at least for the moment. Actually, I'd rather:

1. Leave `cmdhist.lua` as is
1. Load history when the two modes are defined
3. Record history in the `activate` method of each mode (and perhaps dump it there too)
4. Get rid of the `activate` signal for the `input` object, which, if I'm not mistaken, has no other use (that would also require rewriting the `window:activate()` method, but it's no big deal)

### DIFF
--- a/lib/modes.lua
+++ b/lib/modes.lua
@@ -105,9 +105,15 @@ window.add_signal("init", function (w)
             if not w.view.private and hist and hist.items and hist.items[hist.len or -1] ~= text then
                 table.insert(hist.items, text)
                 -- Dump history
-                local f = io.open(luakit.data_dir .. "/" .. mode.name .. "-history", "w")
+                local t = {}
+                for k, v in pairs(modes) do
+                    if v.history then
+                        t[k] = v.history.items
+                    end
+                end
+                local f = io.open(luakit.data_dir .. "/command-history", "w")
                 if f then
-                    f:write(table.concat(hist.items, "\n"))
+                    f:write(lousy.pickle.pickle(t))
                     f:close()
                 end
             end
--- a/lib/modes.lua
+++ b/lib/modes.lua
@@ -101,8 +101,8 @@ window.add_signal("init", function (w)
         if mode and mode.activate then
             local text, hist = input.text, mode.history
             if mode.activate(w, text) == false then return end
-            -- Check if last history item is identical
-            if hist and hist.items and hist.items[hist.len or -1] ~= text then
+            -- Check if last history item should be recorded.
+            if not w.view.private and hist and hist.items and hist.items[hist.len or -1] ~= text then
                 table.insert(hist.items, text)
                 -- Dump history
                 local f = io.open(luakit.data_dir .. "/" .. mode.name .. "-history", "w")
--- a/lib/modes.lua
+++ b/lib/modes.lua
@@ -104,7 +104,7 @@ window.add_signal("init", function (w)
             if mode.activate(w, text) == false then return end
             -- Prevents recording command history if in a private tab, or
             -- recording the `:priv-tabopen` command itself.
-            if w.view.private then return end
+            if w.view and w.view.private then return end
             local hist = mode.history
             -- Check if last history is identical
             if hist and hist.items and hist.items[hist.len or -1] ~= text then
--- a/lib/modes.lua
+++ b/lib/modes.lua
@@ -104,6 +104,12 @@ window.add_signal("init", function (w)
             -- Check if last history item is identical
             if hist and hist.items and hist.items[hist.len or -1] ~= text then
                 table.insert(hist.items, text)
+                -- Dump history
+                local f = io.open(luakit.data_dir .. "/" .. mode.name .. "-history", "w")
+                if f then
+                    f:write(table.concat(hist.items, "\n"))
+                    f:close()
+                end
             end
         end
     end)
--- a/lib/modes.lua
+++ b/lib/modes.lua
@@ -99,10 +99,15 @@ window.add_signal("init", function (w)
     input:add_signal("activate", function ()
         local mode = w.mode
         if mode and mode.activate then
-            local text, hist = input.text, mode.history
+            local text = input.text
+            -- Activates the mode.
             if mode.activate(w, text) == false then return end
-            -- Check if last history item should be recorded.
-            if not w.view.private and hist and hist.items and hist.items[hist.len or -1] ~= text then
+            -- Prevents recording command history if in a private tab, or
+            -- recording the `:priv-tabopen` command itself.
+            if w.view.private then return end
+            local hist = mode.history
+            -- Check if last history is identical
+            if hist and hist.items and hist.items[hist.len or -1] ~= text then
                 table.insert(hist.items, text)
                 -- Dump history
                 local t = {}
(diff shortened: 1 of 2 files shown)


aidanholm:
I've been quietly hoping that someone would implement this, and someone did! :tada: 

With that said, I would prefer that this stored everything in a single pickled table (`lousy.pickle`), rather than using multiple files. Having it edit-unfriendly is a downside, as I can see how it might be useful to edit the history externally, but I think the data dir is already cluttered enough.

zappathustra:
> I've been quietly hoping that someone would implement this, and someone did! 

:)
I've reopened Luakit and wanted my last command so many times, especially while testing!

It's still necessary to save the pickled table to a file, so what you want is reducing multiple files to one? This can be done while keeping it edit-friendly, for instance:

```
[mode:command]
:open foobar
:help
:tabopen barfoo

[mode:lua]
print "hello"
print "world"
```

Also, what do you think of displaying history items that only match the first word of the current line, e.g. if you type `:open`, then only previous commands beginning with `:open` are shown? (This is the way zsh works.) This could be set as an option.

Finally, I think that:
- Given that this is history, users should be made well aware that it is stored somewhere on their computer; and I think it should be possible to disable history (both permanent and from the current session), especially in private tabs.
- Blank lines should be ignored (again, this could be an option).

zappathustra:
Note: `:restart` is not recorded in history. I suppose that's because the associated function kicks in before the history-saver can do anything, but then `:quit` shouldn't be recorded either, yet it is. Anyway that shows I'm not 100% aware of what's going on!

aidanholm:
That sort of plaintext format is also a possibility. pickling is easy, but if you want to go the extra mile I won't complain :)

Do you mean displaying such history items for tab completion?

Hm.. private tabs present an interesting problem, because commands like `:priv-tabopen` may be ran from a non-private tab. It'd be simple enough to have some sort of command blacklist, but typos might slip through... another possible mitigation is to only store the last 100 or so (adjustable with a settings option, ofc)

`:restart` is probably not stored because it just calls `w:restart()`, while `:q` calls `w:close_win()`, which quits only if there are no more windows. Restarting saves the session and then execs the new luakit directly, without closing any windows, and closing the window probably switches modes, which saves the command history. Something like that...

zappathustra:
> Do you mean displaying such history items for tab completion?

No, for history: only browse items that begin like the original command-line, so that if your history is

```
:open foo
:open bar
:tabopen qux
```

then if you type `:open`, only the first two lines will be accessible as previously entered commands (with `<Up>/<Down>`).

> Hm.. private tabs present an interesting problem, because commands like :priv-tabopen may be ran from a non-private tab. It'd be simple enough to have some sort of command blacklist, but typos might slip through... another possible mitigation is to only store the last 100 or so (adjustable with a settings option, ofc)

So there are two problems really:

- (What I meant) Command-line history should be turned off IN a private tab
- (What you mean, I think) `priv-tabopen` should never be recorded in history, even when called from a non-private tab. And of course, there are possible typos, not to mention user-defined commands that create private tabs, but aren't `:private-tab`.

> :restart is probably not stored because it just calls w:restart(), while :q calls w:close_win(), which quits only if there are no more windows. Restarting saves the session and then execs the new luakit directly, without closing any windows, and closing the window probably switches modes, which saves the command history. Something like that...

So I think for the moment I'll just write a special rule for `:restart` :)

aidanholm:
Oh, right, like ctrl-r and ctrl-s for the shell. That sounds good.

For `:restart`, it'd probably be best for the `w:restart()` method to return to normal mode on all windows before exec-ing.

McSinyx:
> Oh, right, like ctrl-r and ctrl-s for the shell.

I think that the exact behavior would be more similiar to Vim command history (search for `^pattern` if shell history supported such a thing)

zappathustra:
> > Oh, right, like ctrl-r and ctrl-s for the shell.

> I think that the exact behavior would be more similiar to Vim command history (search for ^pattern if shell history supported such a thing)

That's it. I'm sorry, but since I set it to default in zsh, I didn't realize it was confusing. But it made me realize that instead of an option (say `history.filter`), there could be two sets of keybindings (one for normal history, another for filtered history, Vim-style) or three (normal, filtered on the first word, filtered on the entire line, as is possible in zsh).

> For :restart, it'd probably be best for the w:restart() method to return to normal mode on all windows before exec-ing.

If I understand how it works (haven't really dived into the code), I'll do that. Otherwise I'll let it to you!

aidanholm:
I think just having it auto-filter by default is enough to begin with: it's pretty easy to hit `ctrl-u` to clear the line, after all.

zappathustra:
Actually, the variants are pretty trivial to implement, but we can begin simply. I didn't know about `ctrl-u`, so I've opened `readline.lua`. Don't you think the modified `cmdhist.lua` could be incorporated there? Also, what about defining command-line bindings (those in `readline.lua`) with `mode.add_binds`? Or did you avoid doing that for a reason?

lib/modes.lua
```diff
@@ -101,8 +101,8 @@ window.add_signal("init", function (w)
         if mode and mode.activate then
             local text, hist = input.text, mode.history
             if mode.activate(w, text) == false then return end
-            -- Check if last history item is identical
-            if hist and hist.items and hist.items[hist.len or -1] ~= text then
+            -- Check if last history item should be recorded.
+            if not w.view.private and hist and hist.items and hist.items[hist.len or -1] ~= text then
```
zappathustra: This filters off both commands entered in a private tab (as expected) but also `:priv-tabopen` itself (unexpected, but I guess that's because we're in a private tab when the history is recorded).

zappathustra:
Neither `w:set_mode()` nor `w:activate()` at the beginning of the `restart` method have any effect: `:restart` is still not registered in history.

lib/modes.lua
aidanholm: oh, that makes sense: could you please add a comment to that effect? also, it may be preferable to just return early if w.view.private is true

lib/modes.lua
```diff
@@ -99,10 +99,15 @@ window.add_signal("init", function (w)
     input:add_signal("activate", function ()
         local mode = w.mode
         if mode and mode.activate then
-            local text, hist = input.text, mode.history
+            local text = input.text
+            -- Activates the mode.
             if mode.activate(w, text) == false then return end
```
zappathustra, line 104: No activable mode (`command`, `lua`, `search` and `completion`) returns `false`, but perhaps testing for it is a mechanism available for other modes (to be invented) that shouldn't be recorded in history?

lib/modes.lua
```diff
@@ -99,11 +99,28 @@ window.add_signal("init", function (w)
...
-            -- Check if last history item is identical
+            -- Prevents recording command history if in a private tab, or
+            -- recording the `:priv-tabopen` command itself.
+            if w.view.private then return end
```
aidanholm: `w.view` can be nil here, so probably `w.view and w.view.private` is needed

aidanholm (commented):
Sorry to have left this one waiting for so long :/ it looks really good to me, just one minor error to fix

aidanholm:
Merged at 3ce519f! Thanks for working on this one :)

PR #661, opened 2018-01-22
added prevent_default handling for dom events

serg-kozhemyakin:
dom event's callback function can set prevent_default to true for its argument and in this case after returning from callback webkit_dom_event_prevent_default will be called for this event.

### DIFF
--- a/extension/clib/dom_element.c
+++ b/extension/clib/dom_element.c
@@ -423,13 +423,13 @@ event_listener_cb(WebKitDOMElement *UNUSED(elem), WebKitDOMEvent *event, gpointe
        stopPropagation */
     lua_pushliteral(L, "cancel");
     lua_rawget(L, -2);
-    if (lua_isboolean(L, -1)) {
-        lua_pushboolean(L, true);
-        if (lua_equal(L, -1, -2)) {
-            webkit_dom_event_stop_propagation(event);
-        }
-        lua_pop(L, 1);
-    }
+    if (lua_toboolean(L, -1)) webkit_dom_event_stop_propagation(event);
+    lua_pop(L, 1);
+    /* also check if prevent_default set to true and if it's set then
+       call webkit_dom_event_prevent_default for event */
+    lua_pushliteral(L, "prevent_default");
+    lua_rawget(L, -2);
+    if (lua_toboolean(L, -1)) webkit_dom_event_prevent_default(event);
     lua_pop(L, 2);
 }
 


extension/clib/dom_element.c
```diff
@@ -430,6 +430,18 @@ event_listener_cb(WebKitDOMElement *UNUSED(elem), WebKitDOMEvent *event, gpointe
         }
         lua_pop(L, 1);
     }
+    lua_pop(L, 1);
+    /* also check if prevent_default set to true and if it's set then
+       call webkit_dom_event_prevent_default for event */
+    lua_pushliteral(L, "prevent_default");
+    lua_rawget(L, -2);
+    if (lua_isboolean(L, -1)) {
```
aidanholm: Should probably be `if (lua_toboolean( ... ))`, which simplifies the code below

aidanholm (changes requested):

extension/clib/dom_element.c
serg-kozhemyakin: hm, if i changing
```c
    lua_pushliteral(L, "prevent_default");
    lua_rawget(L, -2);
    if (lua_isboolean(L, -1)) {
        lua_pushboolean(L, true);
        if (lua_equal(L, -1, -2)) {
            webkit_dom_event_prevent_default(event);
        }
        lua_pop(L, 1);
    }
    lua_pop(L, 2);
```
to
```c
    lua_pushliteral(L, "prevent_default");
    lua_rawget(L, -2);
    if (lua_toboolean(L, -1)) {
        webkit_dom_event_prevent_default(event);
        lua_pop(L, 1);
    }
    lua_pop(L, 1);
```
then webkit process starts crashing after any callback. probably lua_toboolean pops stack itself (don't found info about its behavior in lua doc)

aidanholm: There shouldn't be a `lua_pop()` call inside the if-statement body.

Looks fine when that's fixed: pushliteral+rawget pushes the boolean, toboolean checks it, and the final pop removes it.

serg-kozhemyakin:
updated commit. and also simplified 'cancel' check as well. now looks working so far.

extension/clib/dom_element.c
```diff
@@ -423,12 +423,16 @@ event_listener_cb(WebKitDOMElement *UNUSED(elem), WebKitDOMEvent *event, gpointe
        stopPropagation */
     lua_pushliteral(L, "cancel");
     lua_rawget(L, -2);
-    if (lua_isboolean(L, -1)) {
-        lua_pushboolean(L, true);
-        if (lua_equal(L, -1, -2)) {
-            webkit_dom_event_stop_propagation(event);
-        }
-        lua_pop(L, 1);
+    if (lua_toboolean(L, -1)) {
```
aidanholm: no need for { } here

extension/clib/dom_element.c
```diff
@@ -423,12 +423,16 @@ event_listener_cb(WebKitDOMElement *UNUSED(elem), WebKitDOMEvent *event, gpointe
...
+        webkit_dom_event_stop_propagation(event);
+    }
+    lua_pop(L, 1);
+    /* also check if prevent_default set to true and if it's set then
+       call webkit_dom_event_prevent_default for event */
+    lua_pushliteral(L, "prevent_default");
+    lua_rawget(L, -2);
+    if (lua_toboolean(L, -1)) {
```
aidanholm: ditto

aidanholm (changes requested):
just formatting changes needed

extension/clib/dom_element.c
serg-kozhemyakin: updated

PR #662, opened 2018-01-25
Abbreviate search engine names.

zappathustra:
When you have e.g. `wikipedia` defined as a search engine, it is much more convenient to type only `:open wi something` or `:open wiki something` than the full engine's name.

To do that, I had a loop in `userconf.lua` to create the `w`, `wi`, `wik` ... `wikipedia` search engines. But it is highly unelegant, and it makes the `searchmenu` plugins unusable.

This allows typing only the first letters of a search engine. The minimal number of letters is set with `window.search_engine_min_length`, so a large value can totally disable abbreviations.

### DIFF
--- a/lib/window.lua
+++ b/lib/window.lua
@@ -832,7 +832,7 @@ settings.register_settings({
     },
     ["window.search_engine_min_length"] = {
         type = "number", min = 1,
-        default = 2,
+        default = 100,
         desc = "How many letters make a valid search engine abbreviation.",
     },
     ["window.scroll_step"] = {
--- a/lib/window.lua
+++ b/lib/window.lua
@@ -280,6 +280,16 @@ local init_funcs = {
     end,
 }
 
+local table_keys, filter_array = lousy.util.table.keys, lousy.util.table.filter_array
+local function check_search_engine (str)
+    if not str or #str < settings.get_setting("window.search_engine_min_length") then return end
+    local engines = table_keys(settings.get_setting("window.search_engines"))
+    local t = filter_array(engines, function (_, n)
+        return n:sub(1, #str) == str
+    end)
+    return t[1]
+end
+
 --- Helper functions which operate on the window widgets or structure.
 -- @type {[string]=function}
 -- @readwrite
@@ -581,7 +591,6 @@ _M.methods = {
     -- Intelligent open command which can detect a uri or search argument.
     search_open = function (_, arg)
         local lstring = lousy.util.string
-        local search_engines = settings.get_setting("window.search_engines")
 
         -- Detect blank uris
         if not arg or arg:match("^%s*$") then return settings.get_setting("window.new_tab_page") end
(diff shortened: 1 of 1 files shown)


McSinyx:
I'm personally against this, at least the default of `2`. The feature IMHO should be turned off by default as it will break backward-compatibility. Also are you aware that [DuckDuckGo has bang-shorthand](https://duckduckgo.com/bang_lite.html) for many search engines? e.g. if you set default search engine to DDG, then you can simply `:open !w something`.

zappathustra:
> I'm personally against this, at least the default of 2. The feature IMHO should be turned off by default as it will break backward-compatibility.

You are perfectly right. I set the default as if the feature had always existed, which is a bad assumption. Now it's fixed.

> Also are you aware that DuckDuckGo has bang-shorthand for many search engines? e.g. if you set default search engine to DDG, then you can simply :open !w something.

I didn't know that, and that's cool, especially given the sheer number of engines. But I have things they don't, in particular search engines defined as functions.

McSinyx:
As an user, long as you don't change the experience w/o notifying, I won't complain.

But about the fix, well, I expected something different (or less hacky), like a value to disable the feature, which could be either `0` or `-1` in this case.

zappathustra:
That's what I thought too, but I simplified the inner code rather than the UI. Bad assumption again, I guess :smiley:

aidanholm:
I'm not sure if this feature is a good idea; what problem does it solve? It seems like it'd be easier to just define a wikipedia search engine with a short name. Perhaps search engines should have a name field separate to the shortcut name used in the URL bar?

zappathustra:
I'll admit this may have more to do with some personal muscle memory and the way I use Luakit (and Conkeror before that) than with general utility.

User-defined aliases (e.g. `wi`) would be simpler, but I'm not consistent in what I use (I sometimes type `w`, sometimes `wiki`, and so on... multiply that by the number of search engines...), that's why it's inconvenient to me. But perhaps I should re-educate myself instead of blaming Luakit :)

Your call anyway!

aidanholm:
I would prefer to see this implemented as a plugin, rather than a modification of `window.lua`. This would probably mean exposing `search_open` to customization via some sort of signal; perhaps something like this:
```lua
window.add_signal("search_open", function (args)
    -- args is a table that holds the string arguments
end)
```
This way, successive signal handlers can be chained. Your plugin would become one signal handler that simply expands the first word of the string to the full search engine name.

zappathustra:
I've been thinking about my solution and I realized I unconsciously exaggerated the number of abbreviations I was using: it is true that I may have four or five different ones for `wikipedia`, but not for the other engines. So for the moment I'll make do with aliases. So obviously you can delete this PR.

On the other hand, exposing `search_open` remains a good idea, I think. I'm generally a big fan of such callbacks or hooks, I think it's a great way to make a program customizable. In my opinion, the whole dealing-with-engines stuff should be a plugin exploiting such a signal. The only job of Luakit would be to pass the command line to the signal and expect an URL in return. But I digress :)

aidanholm:
I'm a fan of these kinds of hooks as well. The only drawback is hooks added by multiple plugins; on most hooks, that's never an issue, but sometimes you'll have ordering dependencies, and this search hook is potentially one of them. For example, your search engine expanding hook would need to run before any hooks that looked at the search engine name.

PR #663, opened 2018-01-27
Add gopher:// support

Ygrex:


aidanholm:
This is very cool, and I'm surprised by how little code was necessary.

I haven't looked too closely at the code yet, but I've noticed two issues:
1) some pages at `gopher://gopher.quux.org/1/` don't seem to work, specifically those with spaces/%20 in the URL
2) luakit is blocked while the load is performed. Probably the best way to fix this is by returning from the `and return` signal handler immediately (after URL verification + error page stuff), using non-blocking `socket` calls from a coroutine, and calling `request:finish()` from that coroutine. You'll need to run that coroutine from an idle callback until it completes.

Ygrex:
thank you, fixed both issues

aidanholm:
As I kind of suspected, using idle_cb is better in that the browser isn't blocked, but it's still very laggy. Here's a patch for this PR that uses a timer instead: works much better:
```
diff --git a/lib/gopher.lua b/lib/gopher.lua
index 20b4d89b..2ebabd86 100644
--- a/lib/gopher.lua
+++ b/lib/gopher.lua
@@ -304,6 +304,12 @@ local function net_request(host, port, msg)
     return data
 end
 
+local load_timer = timer{interval = 50}
+local num_loads = 0
+load_timer:add_signal("timeout", function ()
+    if num_loads == 0 then load_timer:stop() end
+end)
+
 webview.add_signal("init", function(view)
     view:add_signal("scheme-request::gopher", function(_, uri, request)
         local url = assert(parse_url(uri))
@@ -311,20 +317,27 @@ webview.add_signal("init", function(view)
         local net = coroutine.wrap(function()
             return net_request(url.host, url.port, msg .. "\r\n")
         end)
-        luakit.idle_add(function()
+        if not load_timer.started then load_timer:start() end
+        num_loads = num_loads + 1
+        local function remove_loader(loader)
+            num_loads = num_loads - 1
+            load_timer:remove_signal("timeout", loader)
+        end
+        local function loader ()
             local status, res = pcall(net)
             if not status then
                 request:finish("Error: " .. tostring(res), "text/plain")
-                return false
+                return remove_loader(loader)
             end
             if not res then
-                return true
+                return
             end
             if not request.finished then
                 request:finish(data_to_browser(res, url))
             end
-            return false
-        end)
+            remove_loader(loader)
+        end
+        load_timer:add_signal("timeout", loader)
     end)
 end)
```
I probably won't enable this plugin by default, as it requires an additional dependency (luasocket); so this plugin will also need to log an error message and do nothing (return an empty module table) if `require("socket")` fails.

It'd be nice to get a callback when data is ready, but according to `https://stackoverflow.com/questions/12889361/lua-sockets-asynchronous-events` we're kind of stuck with polling with luasocket. If glib provides some relevant APIs, wrapping those could be a great improvement. Losing that dep would mean I would be happy to enable this by default as well.

Another issue I've noticed is that stopping the page load isn't handled: webkit handles stopping the URI scheme request, but the luasocket load continues in the background. It'd be preferable to stop the load immediately inside the `"load-status"` handler, but I'd also accept checking for cancellation when the polling function is next called.

lib/gopher.lua
```diff
@@ -0,0 +1,333 @@
...
+
+local socket = require("socket")
+local webview = require("webview")
+
+local _M = {}
+
+luakit.register_scheme("gopher")
+
+local function gophertype_to_icon(gophertype)
+    -- TODO unicode symbols look different
```
aidanholm, line 23: You could use inline svg, if you can find an icon set that you like; unfortunately licensing could be an issue here

lib/gopher.lua
```diff
@@ -0,0 +1,333 @@
...
+    elseif url.gophertype == "9" then
+        mime = "application/octet-stream"
+    elseif url.gophertype == "g" then
+        mime = "image/gif"
+    elseif url.gophertype == "p" then
+        mime = "image/png"
+    elseif url.gophertype == "I" then
+        mime = image_mime_type(url.selector:lower():match("%.(.-)$"))
+    elseif url.gophertype ~= "h" then
+        print("Unsupported item type: '" .. url.gophertype .. "'")
```
aidanholm: should be using msg.error, probably

lib/gopher.lua
```diff
@@ -0,0 +1,333 @@
...
+        port = port,
+        gopher_path = gopher_path,
+        gophertype = gophertype,
+        selector = selector,
+        search = search,
+        gopher_plus_string = gopher_plus_string,
+    }
+end
+
+do
```
aidanholm: testing code here would be better moved to `tests/async/test_gopher.lua`

lib/gopher.lua
```diff
@@ -0,0 +1,333 @@
...
+        coroutine.yield()
+    end
+    return sent
+end
+
+-- non-blocking reading
+local function _net_read_data(conn)
+    local chunks = {}
+    while true do
+        local res, err, last = conn:receive("*a")
```
aidanholm, line 373: why loop if you wait to receive everything? probably this should use a chunk size of 1mb or something

aidanholm (changes requested):

lib/gopher.lua
aidanholm: Images with data URIs would also work

IWriteThings:
+1 on Gopher support. I could REALLY use this. I use Luakit all the time and being able to use it for my Gopher stuff too would be a huge boon.

Thanks!!!!

hippi777:
hi there!
if it will be done, then the wikipedia page could be updated for some random advertising :D 
https://en.wikipedia.org/wiki/Gopher_protocol
i mean the "clienst" on the page...
(just tried 2 figure out what could this be :D )

lib/gopher.lua
```diff
@@ -0,0 +1,443 @@
...
+            )
+        end
+    end
+    return src
+end
+
+local function text_to_html(data)
+    return [[
+        <html>
+        <head>
```
aidanholm, line 127: This should include the title of the page

lib/gopher.lua
```diff
@@ -0,0 +1,443 @@
...
+        mime = "image/gif"
+    elseif url.gophertype == "M" then
+        mime = "application/mbox"
+    elseif url.gophertype == "p" then
+        mime = "image/png"
+    elseif url.gophertype == "I" then
+        mime = image_mime_type(url.selector:lower():match("%.(.-)$"))
+    elseif url.gophertype ~= "h" then
+        msg.error("Unsupported item type: '%s'", url.gophertype)
+        error("Unsupported item type")
```
aidanholm: This should use `error_page.show_error_page()`

lib/gopher.lua
```diff
@@ -0,0 +1,443 @@
...
+        local function loader ()
+            if loads[v] ~= net then
+                remove_loader(v, loader)
+                return
+            end
+            local alive, _ = pcall(function() return v.is_loading end)
+            local status, res = coroutine.resume(net, not alive)
+            if not status then
+                if not request.finished then
+                    request:finish("Error: " .. tostring(res), "text/plain")
```
aidanholm: This should also use `show_error_page()`, after the `request:finish()` call

ghost:
@Ygrex Do you think you can find the time to work on the requested changes? I'm really looking forward to this feature.

Ygrex:
yes, I have time for this, the limiting factor there has been the fact that master stopped working for me (https://github.com/luakit/luakit/issues/681)
that means that even if I make this get merged, I'll not be able to use it myself

thanks for the interest in this feature, I'll continue working on this in a fork

Ygrex:
I failed with SVG iconset, other spotted issues are fixed now

lib/gopher.lua
iamleot: Regarding the **TODO** comment... maybe it just depends on the font?
(or, digging in the gopherspace with the patch applied I have not seen anything
"weird" about the symbols used for gopher Item Types :))

c0dev0id: Hmm, what about not using icons at all? I think they look less clean than text. On my page, I'm just using text and I'm very happy with it: https://codevoid.de (this is the web frontend for my gopherhole)

Ygrex:
there is simpler and cleaner version based on this:
https://github.com/luakit/luakit/pull/884

PR #669, opened 2018-01-31
New setting: window.reuse_new_tab_pages

vlnx:


lib/window.lua
```diff
@@ -441,12 +441,27 @@ _M.methods = {
             view = arg
             local ww = webview.window(view)
             ww:detach_tab(view)
-        else
+            w:attach_tab(view, switch, order)
+        end
+
+        if not view and settings.get_setting("window.reuse_new_tab_pages") then
+            for tabindex, tab in ipairs(w.tabs.children) do
```
aidanholm: `tabindex` is unused and should be just `_`; other than that, everything looks great!

aidanholm (changes requested):

PR #702, opened 2018-04-23
add downloads.db

amosbird:
This commit adds downloads.db to store all finished download items.

### DIFF
--- a/lib/downloads.lua
+++ b/lib/downloads.lua
@@ -19,6 +19,13 @@ local add_binds, add_cmds = modes.add_binds, modes.add_cmds
 
 local _M = {}
 
+--- Path to downloads database.
+-- @readwrite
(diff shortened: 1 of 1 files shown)


lib/downloads.lua
```diff
@@ -147,12 +179,15 @@ function _M.add(uri, opts)
...
             end)
-            --return true
         else
             dd:cancel()
         end
         return true
     end)
+
+    d:add_signal("finished", function(dd)
+                     query_insert:exec{os.time(), dd.uri, dd.destination}
```
aidanholm: indentation

aidanholm (commented):
The biggest concern I have is that this isn't widely useful for the majority of users. It's not even obvious that all downloads are being recorded; it'd be preferable if these downloads were also present in the luakit downloads list.

aidanholm:
Thanks!

PR #709, opened 2018-06-07
New functions for styles module

gleachkr:
This adds:
1. live editing of user styles,
2. toggling of user styles, and
3. creation of a new user style for the current uri.

### DIFF
--- a/lib/styles.lua
+++ b/lib/styles.lua
@@ -413,7 +413,9 @@ add_binds("styles-list", lousy.util.table.join({
             local row = w.menu:get()
             if row and row.stylesheet then
                 local file = luakit.data_dir .. "/styles/" .. row.stylesheet.file
-                editor.edit(file)
+                local guard = {0}
+                _M.watch_styles(guard, file)
+                editor.edit(file, 1, function() guard[1] = nil end)
             end
         end },
 }, menu_binds))
--- a/lib/styles.lua
+++ b/lib/styles.lua
@@ -375,7 +375,7 @@ end
 
 --- Create and immediately edit a new style for the current uri.
 -- @tparam table w The window table for the window providing the uri.
-M.new_style = function (w)
+_M.new_style = function (w)
     -- Create styles directory if it doesn't exist
     local cwd = lfs.currentdir()
     if not lfs.chdir(styles_dir) then
@@ -398,6 +398,19 @@ M.new_style = function (w)
             local guard = {0}
             _M.watch_styles(guard, path)
             editor.edit(path, 2, function() guard[1] = nil end)
+        end
+    end
+    lfs.chdir(cwd)
+end
+
+--- Toggle the enabled status of a style by filename.
+-- @tparam string title the style to toggle.
+_M.toggle_sheet = function(title)
+    for _, stylesheet in ipairs(stylesheets) do
+        if stylesheet.file == title then
+            stylesheet.enabled = not stylesheet.enabled
+            db_set(stylesheet.file, stylesheet.enabled)
+            update_all_stylesheet_applications()
         end
     end
 end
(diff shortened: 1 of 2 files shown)


lib/styles.lua
```diff
@@ -362,6 +362,16 @@ _M.detect_files = function ()
     lfs.chdir(cwd)
 end
 
+--- Watch the styles directory for changes and apply them immediately.
```
gleachkr: This is probably the most adventurous bit of the pull request. Suggestions welcome. It might be nice to have this function return the guard table, or a closure, rather than passing the table in, but for some reason neither of those approaches seem to play nicely with the callback in `luakit.spawn`.

aidanholm: I'm not too happy with polling, although in this case it's not too bad. I'd prefer to see use of GFileMonitor, but that would need to be wrapped to be usable from lua, which could be a fair amount of work.

lib/styles.lua
```diff
@@ -402,7 +455,9 @@ add_binds("styles-list", lousy.util.table.join({
             local row = w.menu:get()
             if row and row.stylesheet then
                 local file = luakit.data_dir .. "/styles/" .. row.stylesheet.file
-                editor.edit(file)
+                local guard = {0}
+                _M.watch_styles(guard)
+                editor.edit(file,1,function() guard[1] = nil end)
```
aidanholm: spaces after commas; there's a few other places with this (minor) problem

lib/styles.lua
gleachkr: Yeah, polling is not ideal. It's sort of a trade-off between this or adding dependencies and/or complexity. It might also be possible to either pull in an off-the-shelf inotify wrapper for lua, or to avoid polling in the presence of something like inotifywatch (falling back on sleep in the case of absence), by adding a little logic to the spawned process. But I thought that the simplicity here might outweigh the inelegance.

lib/styles.lua
gleachkr: Right, will fix.

lib/styles.lua
aidanholm: Unfortunately inotify is not available on the BSDs, so that's why GFileMonitor would be needed as an abstraction.

Maintenance of all the semi-functional C wrappers is becoming enough of a burden that I'm thinking of simply requiring gobject introspection, which would permit accessing GFileMonitor directly from lua. I believe there are already reasonably-high quality wrappers for glib, gio, etc that provide lua-like APIs. But that's a fair bit of work as well.

I think I'll be satisfied with just polling for now, especially since it's not active all the time. If you want to add in something like inotifywatch that'd be welcome, as long as polling is present as a fallback.

gleachkr: OK, I've added the inotifywatch (or actually, inotifywait, which was a little more convenient). There's a ten-second timeout so that it cleans itself up after the editor or luakit exits.

gleachkr: simplified the bash command slightly, and en-quoted the path for bash.

aidanholm: Thanks for the fixes. I have exams coming up, and won't have time to give this further review for another two weeks or so, but overall it looks pretty good.

gleachkr: OK, sounds good. Thanks for the update. Good luck on your exams!

lib/styles.lua
```diff
@@ -370,6 +423,7 @@ add_cmds({
         end },
     { ":styles-list", "List installed userstyles.",
         function (w) w:set_mode("styles-list") end },
+    { ":styles-new", "Create new userstyle for this domain", _M.new_style},
```
aidanholm: the binding description needs a full stop

aidanholm (changes requested):
One tiny nit, and then this is ready for merge!

gleachkr:
OK, should be ready to go!

aidanholm:
Merged! Thanks, and sorry for the wait!

PR #714, opened 2018-08-11
Avoid emiting stale download::status signals.

amosbird:


aidanholm:
Thanks! sorry for the delay

Issue #725, opened 2018-10-06
luakit get stuck and start using all the CPU when closing the last tab


iamleot:
When closing the last tab and when `window.close_with_last_tab` setting is `false` (default) luakit is no longer responsive (despite some events appears when invoked via `--log=DEBUG`) and start using all the CPU.

c0dev0id:
Does this still occur for you people?

I tried to reproduce it with the current luakit and the CPU dropped to zero after closing the last tab.

Video: https://codevoid.de/h/luakit_cpu.mp4

iamleot:
At least for me, yes, I can still reproduce it with current Git HEAD.

PR #726, opened 2018-10-06
Do not destroy() the current view, workaround for issue #725

iamleot:
When just one tab is present it leads to an high CPU usage and luakit completely
stuck (although start luakit via `--log=DEBUG' option shows that it responds to
events).

### DIFF
--- a/lib/window.lua
+++ b/lib/window.lua
@@ -481,7 +481,6 @@ _M.methods = {
         view = view or w.view
         w:emit_signal("close-tab", view)
         w:detach_tab(view, blank_last)
-        view:destroy()
     end,
 
     attach_tab = function (w, view, switch, order)


aidanholm:
Thanks for the PR! I'd prefer to see this special-cased for the particular situation described in #725, however; i.e. don't call `destroy()` for that setting value if the last tab is being closed, but do call it otherwise. Ideally, a timer would be set for a few seconds, after which this tab would be closed.

iamleot:
Hello Aidan,

Aidan Holm writes:
> Thanks for the PR! I'd prefer to see this special-cased for the particular situation described in #725, however; i.e. don't call `destroy()` for that setting value if the last tab is being closed, but do call it otherwise. Ideally, a timer would be set for a few seconds, after which this tab would be closed.

Whoops, nice catch!

Sure, I'll try to adjust the patch as discussed!


Thank you very much for the feedback!

iamleot:
I have just adjusted it to reflect the case by just checking
w.tabs:count() and verified that it correctly works.

Please let me know if it missing something/needs further adjustments!

iamleot:
Leonardo Taccari writes:
> I have just adjusted it to reflect the case by just checking
> w.tabs:count() and verified that it correctly works.
>
> Please let me know if it missing something/needs further adjustments!

Err, no, I'm wrong.

Now it seems that I can randomly get luakit stuck when opening
several tabs and closing one of them, also when there is still more
than one tab.

iamleot:
Leonardo Taccari writes:
> Err, no, I'm wrong.
>
> Now it seems that I can randomly get luakit stuck when opening
> several tabs and closing one of them, also when there is still more
> than one tab.

I think that calling `view:destroy()' can be problematic also when
more than one tab is present.  At least after changing:

    view:destroy()

to:

    if w.tabs:count() > 1 then view:destroy() end

I see the following on pkgsrc-wip^[0] wip/luakit-git^[1] package
where latest HEAD in Git of luakit is built.  And, luakit is
configured with USE_LUAJIT=0 and Lua 5.1 is used instead (if this
can make any difference).

I have reverted the patch on pkgsrc-wip to unconditionally avoid to call
view:destroy() ATM.

Any idea on what's going on and/or any possible idea how to debug
that further?


Thank you!


[0]: https://www.pkgsrc.org/wip/
[1]: http://pkgsrc.se/wip/luakit-git

iamleot:
For consistency I have reverted it to the version used in pkgsrc.

(But if you have any possible idea/patch please let me know, thanks!)

c0dev0id:
Completely different idea:
What about not destroying the webview at all. We could check on "tab close" if it is the last tab and then not close it, but leave it open and navigate to the "new tab" page instead.

This would make the empty space configurable and we would not run into this problem at all as there would always be one webview loaded.

What do you think about this?

c0dev0id:
Fixed with https://github.com/luakit/luakit/commit/9dfaeb8d1f97369722f0cd1cda4d661236e55905

PR #727, opened 2018-10-06
Avoid DESTDIR in buildopts.h definitions

iamleot:
Improve DESTDIR support and use it only in install and deinstall target.
The various paths are used in buildopts.h and there having the
DESTDIR is not desiderable.


iamleot:
Please note that this probably partly overlaps with pull request #549.

c0dev0id (approved):
Test without DESTDIR & PREFIX: ok
Test with DESTDIR & PREFIX: ok
Test with PREFIX: ok
Test with DESTDIR: ok

Also good that the xdg config is placed under DESTDIR.

I'll merge this patch. Thank you!

Issue #77, opened 2012-04-25
World readable cookie.db


mason-larobina:
Find an appropriate way to change the permissions on the `$XDG_{CACHE,CONFIG,DATA}_DIR/luakit` dirs to prevent other users peeking at a users cookies.db and other sensitive data (I.e. form data).

c0dev0id:
Fixed with commits https://github.com/luakit/luakit/commit/092268122af4d512c10d367f0e9da35d2cb8ebb3 and https://github.com/luakit/luakit/commit/33930243cb3d54a45ab47e1ce5bf5747b025e6a0.

PR #728, opened 2018-10-16
Adjust permissions of cookies.db (avoid possible word-readable permissions)

iamleot:
cookies.db by default is created with 0644 permissions leading to
possible sensitive information disclosure on multi user systems.

Pointed out in issue #77 by @mason-larobina, thanks!

Please note that this only applies to cookies.db but other files
should be probably adjusted similarly.

### DIFF
--- a/clib/soup.c
+++ b/clib/soup.c
@@ -24,6 +24,7 @@
 #include "common/signal.h"
 #include "web_context.h"
 
+#include <glib/gstdio.h>
 #include <libsoup/soup.h>
 #include <webkit2/webkit2.h>
 
@@ -100,10 +101,16 @@ static void
 luaH_soup_set_cookies_storage(lua_State *L)
 {
     const gchar *new_path = luaL_checkstring(L, 3);
+    FILE *f;
     if (g_str_equal(new_path, ""))
         luaL_error(L, "cookies_storage cannot be empty");
     g_free(cookies_storage);
     cookies_storage = g_strdup(new_path);
+    
+    if ((f = g_fopen(cookies_storage, "a")) != NULL) {
+        g_chmod(cookies_storage, 0600);
+        fclose(f);
+    }
 
     WebKitWebContext * web_context = web_context_get();
     WebKitCookieManager *cookie_mgr = webkit_web_context_get_cookie_manager(web_context);


c0dev0id:
Hi iamleot,
Interesting - I just checked firefox, chrome, vimb. They all store the cookie database with 0644.

Firefox solves this issues by setting the profile directory to 0700. This is a good way to handle the situation without having to check single files. I think we should set `$XDG_DATA_DIR/luakit` to 0700. Then we don't need to check individual files underneath.

Would you agree to this?

iamleot:
Hello @sdk!

sdk writes:
> [...]
> Interesting - I just checked firefox, chrome, vimb. They all store the cookie database with 0644.

At least in Vimb a similar patch was applied in `master' and new
created configuration files has 0600 permissions (but not on 3.2.0).

> Firefox solves this issues by setting the profile directory to 0700. This is a good way to handle the situation without having to check single files. I think we should set `$XDG_DATA_DIR/luakit` to 0700. Then we don't need to check individual files underneath.
>
> Would you agree to this?

Yes, this will do the trick too!


Thank you for the quick feedback!

c0dev0id:
I would assume that this is as easy as changing the permission bits in `luakit.c -> init_directories`
I've tested this on my fork with commit https://github.com/c0dev0id/luakit/commit/38f2a2ab64c2f2ca63f5c9025759ac7905d1d7e6. And it works fine. I changed all the directory permission (data, cache, config), because I don't see a reason why not. There was also concerns in IRC about the formfiller.lua being readable.

Maybe you want to test this yourself too and maybe @aidanholm wants to comment.

Thanks for bringing this up. I think it's an important topic.

iamleot:
Hello sdk,

sdk writes:
> I would assume that this is as easy as changing the permission bits in `luakit.c -> init_directories`
> I've tested this on my fork with commit https://github.com/c0dev0id/luakit/commit/38f2a2ab64c2f2ca63f5c9025759ac7905d1d7e6. And it works fine. I changed all the directory permission (data, cache, config), because I don't see a reason why not. There was also concerns in IRC about the formfiller.lua being readable.
>
> Maybe you want to test this yourself too and maybe @aidanholm want's comment.
>
> Thanks for bringing this up. I think it's an important topic.

Thank you!  I have tested your patch and it works.

Please note though that possible problems can happens if in the
user configuration soup.cookies_storage of `soup' module is set
outside luakit.{cache,config,data}_dir.

clib/soup.c
```diff
@@ -100,10 +101,16 @@ static void
...
 {
     const gchar *new_path = luaL_checkstring(L, 3);
+    FILE *f;
     if (g_str_equal(new_path, ""))
         luaL_error(L, "cookies_storage cannot be empty");
     g_free(cookies_storage);
     cookies_storage = g_strdup(new_path);
+    
+    if ((f = g_fopen(cookies_storage, "a")) == NULL)
+        luaL_error(L, "Could not open cookies_storage");
```
c0dev0id: No access to the cookie db should not error out. Without one, luakit will fall back to not using a cookie db. If it can't access the file, a warning will already be shown.

clib/soup.c
```diff
@@ -100,10 +101,16 @@ static void
...
+    g_chmod(cookies_storage, 0600);
+    fclose(f);
```
c0dev0id: I would suggest to turn the if-condition around and do something like this:

```
    if ((f = g_fopen(cookies_storage, "a")) != NULL) {
        g_chmod(cookies_storage, 0600);
        fclose(f);
    }
```

This will set the permissions and continues to work without a cookie db if opening the file fails.

c0dev0id:
Merged, thanks!

PR #729, opened 2018-10-16
Add webkit version to :help header

c0dev0id:
This makes it easier for users to find out which webkit version they are using.


aidanholm:
Merged; thanks!

PR #730, opened 2018-10-17
Change default directories to be only accessible by the owner

c0dev0id:
This is an alternative approach to PR https://github.com/luakit/luakit/pull/728

Changing default directory permissions for data / cache / config directory from 771 to 700. This is a better default and will protect possibly sensitive files like browser cache, formfiller.lua and cookie.db.


iamleot:
Please note that this can be still problematic if locations outside data / cache / config directories are used (e.g. in the case of cookies.db by setting soup.cookie_storage).

I think it's probably better to combine both solutions.

PR #731, opened 2018-10-17
 Adding webkit version to "--version" output

c0dev0id:
As we have the webkit version in the `:help` page now, I also added it to the command line `--version` so it's consistent.

### DIFF
--- a/luakit.c
+++ b/luakit.c
@@ -122,6 +122,8 @@ parseopts(int *argc, gchar *argv[], gboolean **nonblock)
     /* print version and exit */
     if (version_only) {
         g_printf("luakit %s\n", VERSION);
+        g_printf("  built with webkit %i.%i.%i ", WEBKIT_MAJOR_VERSION, WEBKIT_MINOR_VERSION, WEBKIT_MICRO_VERSION);
+        g_printf("(installed version: %u.%u.%u)\n", webkit_get_major_version(), webkit_get_minor_version(), webkit_get_micro_version());
         exit(EXIT_SUCCESS);
     }
 


luakit.c
```diff
@@ -121,7 +121,8 @@ parseopts(int *argc, gchar *argv[], gboolean **nonblock)
 
     /* print version and exit */
     if (version_only) {
-        g_printf("luakit %s\n", VERSION);
+        g_printf("luakit %s (using webkit %i.%i.%i)\n", VERSION,
```
aidanholm: It might be better to use `WEBKIT_MAJOR_VERSION` etc. here, as that's what `luakit.webkit_version` returns, and what will be shown in the help page. That's the version of webkit luakit was compiled against, while the functions here return the version that luakit is linked to at runtime.

Of course, it might be desirable to show both somehow.

c0dev0id:
I've added both information. The current output would be:

```
luakit 2017.08.10-371-gcc614da1
  built with webkit 2.22.2 (installed version: 2.22.2)
```

luakit.c
```diff
@@ -122,6 +122,8 @@ parseopts(int *argc, gchar *argv[], gboolean **nonblock)
     /* print version and exit */
     if (version_only) {
         g_printf("luakit %s\n", VERSION);
+        g_printf("  built with webkit %i.%i.%i ", WEBKIT_MAJOR_VERSION, WEBKIT_MINOR_VERSION, WEBKIT_MICRO_VERSION);
+        g_printf("(installed version: %i.%i.%i)\n", webkit_get_major_version(), webkit_get_minor_version(), webkit_get_micro_version());
```
aidanholm:
```suggestion
        g_printf("(installed version: %u.%u.%u)\n", webkit_get_major_version(), webkit_get_minor_version(), webkit_get_micro_version());
```
These functions return unsigned ints.

c0dev0id:
Thanks for the correction. I squashed the commits so it's ready for merge.

Issue #705, opened 2018-05-14
Can't use Javascript 


lzt2002:
I am using Ubuntu 17.10
When I wanted to open a site which uses Javascript, it failed. I can't open that site.
Then I typed /js and found this

![2018-05-14 13-58-41](https://user-images.githubusercontent.com/37217016/39980579-3f1fd894-577f-11e8-948c-c2198d20aa28.png)

Can anybody help me
Thanks

aidanholm:
What version of luakit are you running?

c0dev0id:
Hi @lzt2002 

typing just `:js` will always raise this error. This function evaluates JavaScript. Try calling `:js alert("It works!")` and see if this throws an error.

How does it fail? Blank page? Error?
Is `webview.enable_javascript` checked in `:settings` (assuming you're using the git version)
Any console / `:log` output?

PR #732, opened 2018-10-21
Show nicer error when calling :js without argument

c0dev0id:
This request checks if an argument is present when invoking `:js`. It will present a "nice" error to the user instead of calling eval_js() with an empty argument. This should help with issue https://github.com/luakit/luakit/issues/705

### DIFF
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -489,12 +489,18 @@ modes.add_cmds({
         format = "{uri}",
     }},
     { ":javascript, :js", "Evaluate JavaScript snippet.",
-        function (w, o) w.view:eval_js(o.arg, {
+        function (w, o)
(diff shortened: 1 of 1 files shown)


lib/binds.lua
```diff
@@ -489,12 +489,18 @@ modes.add_cmds({
...
+                w.view:eval_js(o.arg, {
+                        no_return = true,
+                        callback = function (_, err)
+                            w:error(err)
+                        end,
+                    }) 
+                else
+                    w:error("No argument provided")
+                end
+            end },
```
McSinyx: 5 lines above are over indented by 1 level.

c0dev0id:
@McSinyx: Thanks! I saw it too, when I was looking at the github diff.

PR #737, opened 2018-11-16
only signal "mime-type-decision" for a succesful response

yorickhardy:


yorickhardy:
When using privoxy, for example, blocked urls still trigger a mime-type-decision and this seems to always result in the download dialog even though the resource cannot be downloaded.

c0dev0id:
I noticed this as well.The patch looks good to me. I use it for a few days already and haven't noticed any issues.

Thanks!

PR #745, opened 2019-02-11
Fix tablist not hiding in fullscreen

necauqua:
It was bothering me for a long time that the tablist is visible when I open a YouTube video in fullscreen, so I finally fixed it with this in my `userconf.lua`:
```lua
window.add_signal('init', function(w)
    -- fix taglist not hiding when e.g. a youtube video is in fullscreen
    w.win:add_signal('property::fullscreen', function(win)
        w.tablist.widget.visible = not win.fullscreen
    end)
end)
```
But just after I dit that I've noticed the init function [hide_ui_on_fullscreen](https://github.com/luakit/luakit/blob/develop/lib/window.lua#L269) which does what I need except that it is most likely broken.
So here is a fix :)

### DIFF
--- a/lib/window.lua
+++ b/lib/window.lua
@@ -269,7 +269,7 @@ local init_funcs = {
     hide_ui_on_fullscreen = function (w)
         w.win:add_signal("property::fullscreen", function (win)
             w:update_sbar_visibility()
-            w.tablist.visible = not win.fullscreen
+            w.tablist.widget.visible = not win.fullscreen
         end)
     end,
 


aidanholm:
Thanks for the PR! Unfortunately, however, your fix doesn't take the `tablist.always_visible` setting into account, and the needed changes were simple enough that I figured I'd just go ahead and push it myself: it's now fixed on `develop`. Thanks for bringing this to my attention though!

necauqua:
So my fix will remain, because I have always_visible on (so that one tab is still visible) but I want it to hide in fullscreen.
Maybe we need more choises in config?

edit:
Well as far as I understood from your commit, my above statement is not the case (maybe)
I will respond (or not, if everything is fine) later, after I'll have that version and check that
edit2:
Yes, everything is fine now, thanks)

aidanholm:
Glad it's working, and thanks again! :)

Issue #749, opened 2019-02-15
Is wayland supported?


annacrombie:
## I'm submitting a…
Feature Request / Question
## Current Behavior
I'm getting compile errors building against Gtk and WebKit compiled without X and with wayland,
I suspect this is the cause.  The errors are along the lines of
```
widgets/socket.c:27:15: error: unknown type name 'GtkSocket'; did you mean 'AtkSocket'?
```
and
```
widgets/socket.c:68:18: error: 'Window' undeclared (first use in this function); did you mean 'AtkWindow'?
```
## Expected Behavior
Luakit compiles

## Reproduction Instructions
Compile luakit against libraries built without the X use flag, replacing it with wayland

## Environment
Linux Distribution & Version: Gentoo, fresh install

I mainly want confirmation that this is indeed a issue with luakit not supporting wayland, not something else.  And if luakit doesn't support wayland, are there any plans to do so?

hippi777:
hi there :)

wayland is compatible with x but i dunno how, probably its about compiling against x and running in wayland x emulator... otherwise i believe lk and webkitgtk are not written for wayland, but others will tell u if my assumption was wrong

aidanholm:
@annacrombie Hi! Thanks for the report.

A quick google suggests that GtkSocket is not compatible with Wayland, and that some other projects have run into this issue. GtkSocket is only used to provide the socket widget, which is not used at all by the default setup.

You might try removing `widgets/socket.c` and then removing the lines that reference sockets from `clib/widget.h` and `clib/widget.c`. If that fixes Wayland builds, I'd be happy to consider removing the socket widget completely; it's very niche functionality.

annacrombie:
@szbnwer webkitgtk works with wayland as far as I know
@aidanholm ok, I will try this!

annacrombie:
@aidanholm Was able to get luakit to build by following your suggestions, as well as removing one other include statement, which seems to have no effect.  I opened a pull request

hippi777:
@annacrombie thx! good to know. :D

PR #751, opened 2019-02-18
Make luakit compatible with wayland

annacrombie:
With these changes I am able to get luakit running natively on wayland (I made this pull request from luakit on wayland).  I am running sway without an X fallback enabled.

Based on: https://github.com/luakit/luakit/issues/749


aidanholm:
Thanks for the PR!

I noticed that there are some extra merge commits; it'd be preferable if these weren't included in the PR. Additionally, I'd prefer it if you could squash the three commits removing the socket widget, and perhaps add a short explanation to the commit description about the reason for doing so (wayland compatibility).

Thanks!

annacrombie:
Hey, I was trying to do the changes you requested but accidentally messed up my git history pretty bad.  I figured since there aren't many actual changes it would just be easier for me to make the commits on a fresh fork so I closed this and opened #753

PR #753, opened 2019-02-20
Make luakit compatible with wayland

annacrombie:
With these changes I am able to get luakit running natively on wayland (I made this pull request from luakit on wayland). I am running sway without an X fallback enabled.

Based on: #749


widgets/window.c
```diff
@@ -18,7 +18,6 @@
  *
  */
 
-#include <gdk/gdkx.h>
```
aidanholm: If this is built with X11, `#include <gdk/gdkx.h>` is still needed. So this should be wrapped with `#ifdef GDK_WINDOWING_X11` and `#endif` instead of being removed.

aidanholm (commented):
Thanks for the new PR! And sorry for the slow response.

Looks good; one minor fix is needed, however.

aidanholm:
Merged! Thanks :)

PR #756, opened 2019-02-25
Remove default shift-d bind for session deleting

bodqhrohro:
It's an easy to mistype combination, so I think it should not have such a dangerous binding by default. `:q` and `ZQ` are enough already, some more complicated replacement like `ctrl-shift-q` may be added too.


aidanholm:
Thanks; merged! Yeah, good thinking; this one's bitten me too..

PR #778, opened 2019-04-17
improve robustness of session recovering

bodqhrohro:
Using RAM as a temporary storage for any data that should be kept on a non-volatile medium, is dangerous, because an accident (process crash, system crash, power outage) can make these data lost. A good practice is to rewrite/delete the old version of data only when the new version is completely dumped onto the medium.

This commit prevents data corruption in case of crashes: 
* during update of the `recovery_session` file;
* right after the start, when the `session` file is already loaded, but the recovery saving timer was not triggered yet.

### DIFF
--- a/lib/session.lua
+++ b/lib/session.lua
@@ -76,9 +76,11 @@ _M.save = function (file)
     state = istate
 
     if #state > 0 then
-        local fh = io.open(file, "wb")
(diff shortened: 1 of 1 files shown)


aidanholm:
Great! Thanks :)

PR #779, opened 2019-04-18
command :save, to save the complete page as a single mhtml file

sgerwk:
This is different than :dump in that it saves the whole page,not only its html.

I haven't found a simple way to display the result of the operation in the status bar. This is w:notify(...) in lua, but the webkit function for saving to file is asynchronous. Completion fires a callback (a c function) where the result of the operation can be retrieved. w:notify(...) should be called from there.

### DIFF
--- a/lib/binds.lua
+++ b/lib/binds.lua
@@ -12,6 +12,7 @@
 local _M = {}
 
 local window = require("window")
+local webview = require("webview")
 local taborder = require("taborder")
 local settings = require("settings")
 
@@ -566,6 +567,20 @@ end
                     w:notify("Dumped HTML to: " .. file)
                 end)
                 luakit.idle_add(function () coroutine.resume(co) end)
+            end
+        end },
+
+    { ":save", "Save page as shown to file,",
+        function (w, o)
+            local fname = string.gsub(w.win.title, '[^%w%.%-]', '_')..'.mhtml' -- sanitize filename
+            local file = o.arg or luakit.save_file("Save file", w.win, xdg.download_dir or '.', fname)
+            if file then
+                local view = w.view
+                view:add_signal("save-finished", function(v, file, err)
+                    local ww = webview.window(v)
+                    ww:notify(err or ("Saved to: " .. file))
+                end)
+                view:save(file)
             end
         end },
 })
--- a/widgets/webview.c
+++ b/widgets/webview.c
@@ -189,6 +189,45 @@ luaH_webview_load_string(lua_State *L)
     return 0;
 }
 
+struct save_cb_s {
+    const gchar *filename;
+    widget_t *window;
+};
+
+static void
+save_cb(GObject *o, GAsyncResult *res, gpointer user_data) {
+    WebKitWebView *view = (WebKitWebView *) o;
+    struct save_cb_s *scbs = (struct save_cb_s *) user_data;
+    lua_State *L = common.L;
+    GError *err = NULL;
+    gboolean result;
+
+    result = webkit_web_view_save_to_file_finish(view, res, &err);
+
+    luaH_object_push(L, scbs->window->ref);
+    lua_pushstring(L, scbs->filename);
+    if (result)
+        lua_pushnil(L);
+    else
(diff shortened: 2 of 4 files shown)


aidanholm:
Looks good!

The usual way to allow showing a notification is to emit a lua signal from the C callback; lua code can then attach a signal handler; something like this:
```c
luaH_object_push(L, w->ref);
lua_pushstring(L, name);
luaH_object_emit_signal(L, -2, "save-finished", 1, 0);
lua_pop(L, 1);
```

sgerwk:
Thanks!

It should go now. The only doubt I have is the use of variable "w" inside the signal handler without passing it. It seems to work though.

It the code is fine I'll update the man page.

widgets/webview.c
```diff
@@ -189,6 +189,42 @@ luaH_webview_load_string(lua_State *L)
...
+    luaH_object_emit_signal(L, -3, "save-finished", 2, 0);
+    lua_pop(L, 1);
+
+    free(scbs);
+}
+
+static gint
+luaH_webview_save(lua_State *L)
+{
+    struct save_cb_s *scbs = malloc(sizeof(struct save_cb_s));
```
aidanholm: Use `g_new0` here instead of malloc; it's slightly clearer and it checks for allocation failure automatically.

widgets/webview.c
```diff
@@ -189,6 +189,42 @@ luaH_webview_load_string(lua_State *L)
...
+    struct save_cb_s *scbs = (struct save_cb_s *) user_data;
+    lua_State *L = common.L;
+    GError *err = NULL;
+    gboolean result;
+
+    result = webkit_web_view_save_to_file_finish(view, res, &err);
+
+    luaH_object_push(L, scbs->window->ref);
+    lua_pushstring(L, scbs->filename);
+    lua_pushstring(L, result == FALSE ? err->message : "");
```
aidanholm: If there's no error, `nil` should be pushed instead of an empty string.

lib/binds.lua
```diff
@@ -568,6 +568,23 @@ end
...
         end },
+
+    { ":save", "Save page as shown to file,",
+        function (w, o)
+            local fname = string.gsub(w.win.title, '[^%w%.%-]', '_')..'.mhtml' -- sanitize filename
+            local file = o.arg or luakit.save_file("Save file", w.win, xdg.download_dir or '.', fname)
+            if file then
+                local view = w.view
+                view:add_signal("save-finished", function(v, file, err)
+                    if err ~= "" then
```
aidanholm: You should retrieve the current window for the webview inside the callback, like this:
```lua
local ww = webview.window(view)
```
Then use `ww` instead of `w` to notify. This is because the view could switch windows before the save completes (e.g. the tab is detached into a new window); without this change, the notification would show on its old window.

lib/binds.lua
```diff
@@ -568,6 +568,23 @@ end
...
+                        w:notify(err)
+                    else
+                        w:notify("Saved to: " .. file)
```
aidanholm: This block could be:
```lua
local ww = webview.window(view)
ww:notify(err or ("Saved to: " .. file))
```

widgets/webview.c
```diff
@@ -189,6 +189,42 @@ luaH_webview_load_string(lua_State *L)
...
+    luaH_object_emit_signal(L, -3, "save-finished", 2, 0);
+    lua_pop(L, 1);
+
+    free(scbs);
```
aidanholm: Use `g_free` here.

aidanholm (commented):
Looks good! Just a few minor changes needed

lib/binds.lua
sgerwk: I was indeed wondering how to do that. I believe ``webview.window(v)`` is slightly better.

sgerwk:
Would you prefer the changes to be stashed as a single commit?

aidanholm:
Yeah that'd probably be best; thanks!

aidanholm:
Thanks! sorry for the delay on merging this one

PR #792, opened 2019-06-18
Improve performance of lousy.pickle

bodqhrohro:
The original implementation spawned a lot of intermediate concatenated strings, which is terrible in Lua: both for CPU usage, as a string needs to be copied into a new string, and for RAM usage, as old strings keep in memory regardless of their scope until GC destroys them.

With hundreds of tabs, this led to sudden RAM usage quakes during session saving. So, after trying to mitigate this with putting LuaKit in a cgroup, and with tuning the Lua GC to run more often (and to slow up the process even more), I have rewritten the serializer. Enjoy ;-)

### DIFF
--- a/lib/lousy/pickle.lua
+++ b/lib/lousy/pickle.lua
@@ -22,19 +22,19 @@ function Pickle:pickle_(root)
     self._refToTable = {}
     local savecount = 0
     self:ref_(root)
-    local s = ""
+    local buf = {}
 
     while table.getn(self._refToTable) > savecount do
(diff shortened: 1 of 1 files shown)


aidanholm:
This is wonderful, thanks! How did you manage to track the cause of the RAM issues to the serialization code?

bodqhrohro:
@aidanholm just injected prints of `collectgarbage("count")` into the session module, then bisected and bisected until the culprit function was found. Brute, but effective.

Issue #784, opened 2019-05-02
RAM usage quake on upsizing the window


bodqhrohro:
## I'm submitting a…
Bug Report

## Current Behavior
When I upsize (f.e., maximize) the LuaKit window, I get a RAM usage quake of `luakit` process: it grows to more than 1 GB for a while, than falls back to normal.

## Expected Behavior
No abrupt changes of RAM usage.

## Reproduction Instructions
This seems to be related to the large number of tabs or the large session again. I just tried to spawn 64 empty tabs in a clean session, and couldn't reproduce such a quake. However, the stable RAM usage of `luakit` process differs depending on the window size:
* quarter of screen: 105 MB;
* half of screen: 168 MB;
* full screen: 283 MB.

I have favicons and vertical tabs enabled, and a compositing WM (Compiz), if matters.

## Environment

Linux Distribution & Version: Debian 10.
Output of `luakit --version`: luakit 2.0.0
  built with webkit 2.22.6 (installed version: 2.24.1)

bodqhrohro:
This issue is reproducable with disabled favicons too.

Also, sometimes the memory usage grows because of buffers shared with Xorg. They all have size of 33554432 bytes, and their number may grow up to.tens. Is this a leak or do they have some purpose?

aidanholm:
That's odd. I can't reproduce this one at all. It's most likely webkit2gtk doing something strange, since luakit doesn't have any gradual garbage collection like the falling back process you describe. It's possible that it could be caused by luakit via unusual use of the webkit APIs, but there's nothing too funny going on in the main process so I think that's probably unlikely.

I would guess that your 32MB buffers shared with xorg are used during rendering, and 32MB is a somewhat common shared memory size limit on some distros. I doubt those are a leak.

Without some profiling of where allocations are happening, e.g. using massif or heaptrack, it's hard to say why this is happening.

bodqhrohro:
https://0x0.st/zAYn.gz

`heaptrack_gui` shows very strange results for me. All of peaks are less than 100 MB, though the resident memory usage have grown up to 1.3 GB during this session. Maybe the memory chunks are allocated and freed so often that the kernel cannot keep up and defragment the memory in time? Also, it does not show any symbols, though I installed `dbgsym` packages for all of `luakit`, `libcairo2`, `libglib2.0` and `libgtk-3`; tried to run `heaptrack` with and without the `-d` flag, but symbols do not show up in both cases.

bodqhrohro:
[...]

PR #798, opened 2019-07-31
Show WebView widgets only when switching to them

bodqhrohro:
This resolves the annoying https://github.com/luakit/luakit/issues/784 issue. As I suspected, a new WebView is spawned for **every** tab, even for those not loaded yet. Moreover, they are shown right after creation, which makes no sense for hidden widgets. This explains pretty well why all my plenty of non-loaded tabs were re-rendered behind the scenes and occupied tons of RAM for that.

After some more digging, I found the [possible reason](https://developer.gnome.org/gtk3/stable/GtkNotebook.html#gtk-notebook-set-current-page) for such behaviour:
> Note that due to historical reasons, GtkNotebook refuses to switch to a page unless the child widget is visible. Therefore, it is recommended to show child widgets before adding them to a notebook.

So, they should be shown already to be shown, huh. Though, nothing prevents it to be shown right before the switch, which my patch exactly does.

Now the browser performs well for me, but there may be some side effects.

### DIFF
--- a/widgets/notebook.c
+++ b/widgets/notebook.c
@@ -127,6 +127,8 @@ luaH_notebook_switch(lua_State *L)
     gint i = luaL_checknumber(L, 2);
     /* correct index */
     if (i != -1) i--;
+    GtkWidget* page = gtk_notebook_get_nth_page(GTK_NOTEBOOK(w->widget), i);
+    gtk_widget_show(GTK_WIDGET(page));
     gtk_notebook_set_current_page(GTK_NOTEBOOK(w->widget), i);
     lua_pushnumber(L, gtk_notebook_get_current_page(GTK_NOTEBOOK(w->widget)));
     return 1;
--- a/widgets/notebook.c
+++ b/widgets/notebook.c
@@ -128,7 +128,7 @@ luaH_notebook_switch(lua_State *L)
     /* correct index */
     if (i != -1) i--;
     GtkWidget* page = gtk_notebook_get_nth_page(GTK_NOTEBOOK(w->widget), i);
-    gtk_widget_show(GTK_WIDGET(page));
+    gtk_widget_show(page);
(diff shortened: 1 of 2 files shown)


widgets/notebook.c
```diff
@@ -127,6 +127,8 @@ luaH_notebook_switch(lua_State *L)
     gint i = luaL_checknumber(L, 2);
     /* correct index */
     if (i != -1) i--;
+    GtkWidget* page = gtk_notebook_get_nth_page(GTK_NOTEBOOK(w->widget), i);
+    gtk_widget_show(GTK_WIDGET(page));
```
aidanholm: I don't think the `GTK_WIDGET` macro is necessary if you pass in a `GtkWidget*`

widgets/webview.c
```diff
@@ -1438,7 +1438,7 @@ widget_webview(lua_State *L, widget_t *w, luakit_token_t UNUSED(token))
       NULL);
 
     /* show widgets */
-    gtk_widget_show(GTK_WIDGET(d->view));
+    //gtk_widget_show(GTK_WIDGET(d->view));
```
aidanholm: I'd prefer to remove this line completely, as well as the 'show widgets' comment and the blank line between this and the `return w` statement.

aidanholm (changes requested):
Yeah, all tabs are backed by a webview because during normal usage, all tabs in a session would be visited multiple times. Nowadays, luakit can handle many more tabs, and it's more common to see large sessions where you might only look at 10-20% of the open tabs. Lazily showing webviews should be a good start; I believe there's an issue open for eventually unloading tabs that aren't seen very often, but that'll be more difficult to implement properly.

Thanks for finding and fixing this; the code looks good; just a few style fixups needed.

aidanholm:
Merged; thanks!

PR #810, opened 2019-10-16
Disable styling for source code

zappathustra:
When using styles, `view-source:` pages get styled too, which is inconvenient. This disables that.

Next step is independent styling of `view-source:` pages. They already contain CSS, but the html code is not properly tagged so nothing happens.


aidanholm:
Thanks! Manually merged at dd6281a1.

PR #818, opened 2019-12-19
Clear website data

ulrikdem:
This PR adds the `clear_data` module, which provides 2 commands:
* `:clear-data` opens a menu showing the information from the `luakit.website_data` API, and allows clearing the data of {all types,a specific type} for {all domains,a specific domain}.
* `:clear-favicon-db` clears the favicon cache database.

This is supported by 2 new functions in the `luakit` module:
* `luakit.website_data.clear()`, which wraps `webkit_website_data_manager_clear()`.
* `luakit.clear_favicon_database()`, which wraps `webkit_favicon_database_clear()`.

### DIFF
--- a/clib/luakit.c
+++ b/clib/luakit.c
@@ -694,6 +694,15 @@ luaH_luakit_wch_upper(lua_State *L)
 }
 
 static gint
+luaH_luakit_clear_favicon_database(lua_State *UNUSED(L))
+{
+    WebKitWebContext *ctx = web_context_get();
+    WebKitFaviconDatabase *fdb = webkit_web_context_get_favicon_database(ctx);
+    webkit_favicon_database_clear(fdb);
+    return 0;
+}
+
+static gint
 luaH_luakit_push_install_paths_table(lua_State *L)
 {
     lua_createtable(L, 0, 6);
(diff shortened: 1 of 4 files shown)


c0dev0id:
This is giving the user a lot of choices and potentally a very long list of URLs.

What do you think about a direct command scheme like `:clear-[what] [where]`
While `[what]` can be "cache, cookies" and `[where]` can be "all, current, \<searchstring\>".

I would separate cache and cookies. Most of the time I want to delete cookies, but leave the cache alone.

So `:clear-cookies google` would delete all cookies for domains matching \*google\*.
`:clear-cookies current` would delete cookies for the page opened in the current tab.
`:clear-cookies all` would delete all cookies for all urls.
`:clear-cookies` could open a menu similiar to the one you already have.

Same concept for clear-cache.

Is there a need for being able to just flush the favicon cache? Shouldn't this be part of `:clear-cache`?

I could imagine that this could be extended with `:clear-history <searchstring>`

What do you think about this?

ulrikdem:
I originally wrote this for my own use, but when I realized it required C changes and would be useful for others, I decided to make this PR (if at least the C changes - ed143a7, de281be and 6a1f4f5 - are merged, it should be able to create any custom version of this module in Lua). The design was based on exposing the functionality of the WebKit API as directly as possible, which is why there are so many options.

This is also why `:clear-favicon-db` is a separate command (I added it as a bit of an afterthough, since it cannot be cleared with the main website data API). It would be nice if it used the same command as everything else, but it would have to be done carefully to avoid making it feel too inconsistent, since it cannot be cleared per domain.

I agree it would be convenient to be able to select a domain (or the current domain) without having to scroll through the menu. I would suggest something like `:clear-data [where] [what]`, where both parameters are optional and support completion. If the `[what]` is omitted, it still shows the menu, limited to the domains matching `[where]` (or all domains if also omitted).

It may be reasonable to group the data types into categories, like "cache", including `memory_cache`, `disk_cache`, `offline_application_cache` and `hsts_cache`, and "storage", including `session_storage`, `local_storage`, `websql_databases`, `indexeddb_databases`, `plugin_data`, `cookies` and `device_id_hash_salt` (although I would have to research some of the types to be certain if this grouping is appropriate). I wouldn't like to add these in addition to the existing types, since that would result in even more options, with relationships that must be documented and remembered. Having these instead of the existing types might make sense, in which case I like the idea of separate `:clear-cache` and `:clear-storage` commands. I would also like a command like `:clear-cache-and-storage` (probably with a different name). That would probably be enough for my uses, but some users may want more fine-grained control. Maybe we could make the categories configurable, but I'm not sure whether the categories are really necessary.

I will start working on the changes when I have some time.

c0dev0id:
Thank you for the explanation. I'm using this functionality for a while now and I got used to it.

I'm not a big fan of my initial idea any longer as I realized how many data storage options webkit really has. I would assume that most of the time people would want to delete everything, or everything on a specific page. Or a certain type for all pages. At least those are my use-cases.

The only thing that bugs me all the time is that `:clear-data goog<tab>` does not work like in `:open goog<tab>` where the menu shows a filtered list already.

I'd be okay to merge this PR as it is right now and maybe work or refining the menu handling later on.
Are you ok with this @aidanholm?

ulrikdem:
I still think the ideas discussed above, about adding arguments that support completion, would be an improvement. However, I am currently not really motivated to implement it, since I don't use this module very often, and when I do, the current functionality works well enough. So if anyone wants to work on this in the future, feel free to do so.

Issue #665, opened 2018-01-27
Userstyle parser requires space before document subrules


Kyuuhachi:
For example, the following dummy style causes Luakit to crash on startup: `@-moz-documentdomain(""),domain(""){}`. I've temporarily fixed this by locally changing the `%s+` in `styles.lua:239` to `%s*`, but that accidentally lets through `@-moz-documentdomain("")`, which isn't optimal.

n73red:
@-moz-document regexp('http://www\\.example\\.(com|de)/images/.*'), regexp('(?!http://www\\.example\\.com).*'), regexp('http://www\\.example\\.com/(?!members).*'){}

taobert:
Thanks for accepting the PR..

As an addendum, this issue highlights an additional issue, that the style parsing doesn't provide very good error messages when something is (considered) ill-formed. Ideally, there'd be an explicit description of why the parser baulked, rather than just a lua stacktrace.

PR #819, opened 2020-01-10
Correct @-moz-document parsing; resolves issue #665.

taobert:
This corrects #665. The original fix by @Caagr98, and an extra end-of-token check when parsing '@-moz-document' to ensure it's well-formedness.


aidanholm:
Thanks for the PR! :)

Issue #820, opened 2020-01-11
tablist should be optional


taobert:
## I'm submitting a…

Feature Request.

## Current Behavior

Currently the tablist is optional (via the boolean tablist.always_visible setting) when a window has only one tab. If there is more than one tab, the tablist is always displayed.

## Expected Behavior

I would like tthe option of hiding the tablist, even when multiple tabs are present.
I've added a tablist.visibility enum setting which allows for displaying the tablist Always, Never, or only if there's a sole tab.

## Environment

Linux Distribution & Version:
Output of `luakit --version`:

PR #821, opened 2020-01-11
Depricate tablist.always_visible in favour of tablist.visibility. Implements issue #820.

taobert:


lib/lousy/widget/tablist.lua
```diff
@@ -85,10 +85,11 @@ end
 
 local function update_tablist_visibility(tlist)
     if not data[tlist].notebook then return end -- switching notebook
-    if not data[tlist].visible then
-        tlist.widget.visible = false
```
aidanholm: Why the removal of this initial condition here?

lib/lousy/widget/tablist.lua
```diff
@@ -253,12 +254,23 @@ settings.register_settings({
         type = "boolean",
         default = false,
         domain_specific = false,
-        desc = "Whether the tab list should be visible with only a single tab open.",
+        desc = "This is deprecated in favour of tablist.visibility.",
```
aidanholm: I think we should still document whether this option has any effect, and what effect it still has.

aidanholm (commented):
Thanks for the PR!

lib/lousy/widget/tablist.lua
taobert: An excellent question.
I suspect an error on my part. I implemented this for myself long ago but never got round to submitting it back. The initial condition was added early last year, and i may have deleted it during some ham-fisted hand merging. I'll put it back. Thankyou for catching it.

lib/lousy/widget/tablist.lua
taobert: I'll combine the old and new strings, and add a warning that it may be removed in future.

lib/lousy/widget/tablist.lua
taobert: I've made the changes i suggested to my branch.
I'm not sure whether you get notified when that happens, or whether i need to comment again here.

Let me know if there are any other changes you'd like made.

aidanholm:
Thanks! Squashed and merged manually at 2a03b8a1.

Issue #757, opened 2019-02-27
Misleading hints in RTL mode


bodqhrohro:
## I'm submitting a…
Bug Report

## Current Behavior
I accidentally reached an RTL site. Wanted to switch to the English version, but that turned out to be not so simple!
All hints:
![2019-02-27-203456_1366x768_scrot](https://user-images.githubusercontent.com/2134486/53513955-372f5180-3acf-11e9-842b-154239251aec.png)
Hints after pressing `a`:
![2019-02-27-203540_1366x768_scrot](https://user-images.githubusercontent.com/2134486/53513985-5201c600-3acf-11e9-90c8-33c467448ce4.png)

Actually, this is expected behaviour for RTL texts: inclusions of LTR symbols are displayed LTR, but the whole order of a text is still RTL. That's why the semicolon appears after. But this is misleading in this case, as my hintset does not contain RTL symbols, and the hints are separate pieces of text, not parts of an RTL text.

## Expected Behavior

Symbols of a hint with an LTR-only or RTL-only hintset should always appear in the order they will be pressed in: either always LTR or always RTL correspondingly, or based on the page mode: LTR on LTR pages and RTL on RTL pages.

## Reproduction Instructions
http://sdio4lor.servebeer.com/

Configuration:
```
follow.pattern_maker = follow.pattern_styles.match_label

local select = require "select"
select.label_maker = function ()
    local chars = charset("asdfghjkl;")
    return trim(sort(reverse(chars)))
end
```

## Environment

Linux Distribution & Version: Debian 10
Output of `luakit --version`: luakit 2.0.0
  built with webkit 2.22.5 (installed version: 2.22.5)

taobert:
It seems that text direction uses a number of cues,
including the stated direction of the page, and the character set being used.
The character set is more important than the stated direction,
so even if a site is RTL, the Latin characters in the hints are shown LTR.
However, as punctuation characters could also be used in RTL languages,
the semicolon follows the stated direction of the page.

Adding "direction: ltr;" to the .hint_label style overrides the page's direction
and causes the semicolon to be written/interpreted LTR.
Likewise, setting "direction: rtl;" causes the semicolon to be processed RTL even on LTR pages.
It doesn't, however, override Latin characters being processed LTR.

The default hintset is all digits and so doesn't suffer this problem.
I feel that the documentation/source comments should be extended to mention this issue,
and suggest that if a user includes punctuation characters in the hint charset,
[...]

PR #822, opened 2020-01-15
Add explanation/recomendation about hint direction. Explains issue #757.

taobert:
As i mentioned in my comment on issue #757, i feel that documentation is the correct soluton to close the issue.


aidanholm:
Thanks for the PR!

PR #827, opened 2020-03-25
Use correct glib version macros

jvvv:
Commit 79fd45c used the version macros for gnu libc to test for glib version.


aidanholm:
thanks!

PR #830, opened 2020-04-09
Honor window.new_tab_page for cases where tab is opened without uri set

### LINKED ISSUE #752: :tabopen without a URL provides about:blank instead of luakit://newtab
## I'm submitting a…

Bug Report

## Current Behavior

If I press `t` or `o` and hit enter, I get a new `about:blank` page instead of the `luakit://newtab` page.

Closing the last tab with `d` shows luakit://newtab just fine.

`window.new_tab_page` is set to `luakit://newtab`

Watch:
take2

## Expected Behavior

Opening a new, blank tab should open `luakit://newtab`

## Reproduction Instructions

Use stock luakit with the newtab_chrome module loaded. Open a new tab without a URL.

## Environment

Linux Distribution & Version: Arch Linux
Output of `luakit --version`: luakit 2.1-3-g06f8b7ab built with webkit 2.22.6 (installed version: 2.22.6)

craftyguy:
This sets the tab uri to window.new_tab_page when there is no uri passed
to new_tab (e.g. when calling :open or :tabopen without any parameters).
It also uses new_tab_page when restoring sessions with empty tabs.

Fixes #752

### DIFF
--- a/lib/session.lua
+++ b/lib/session.lua
@@ -128,10 +128,10 @@ local restore_file = function (file, delete)
         for _, item in ipairs(win.open) do
             local v
             if not w then
-                w = window.new({"about:blank"})
+                w = window.new({settings.get_setting("window.new_tab_page")})
                 v = w.view
             else
-                v = w:new_tab("about:blank", { switch = item.current })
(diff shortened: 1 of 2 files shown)


aidanholm:
thanks for the fix!

Issue #583, opened 2017-11-06
bookmarks: use existing tab for :bookmark, and don't add dups


aidanholm:
- If there's already a `luakit://bookmarks/` tab open, that should be used instead of opening a new one
- If there's already a bookmark for the given URL, `:bookmark` should edit that bookmark instead

PR #831, opened 2020-04-10
bookmarks_chrome: switch to bookmark manager tab if already opened

craftyguy:
This implements part of #583 by switching to a tab with the bookmark
manager if calling ":bookmarks" when the manager is already open.


aidanholm:
thanks for the pr!

PR #832, opened 2020-04-14
Fix GLib log messages

ulrikdem:
For GLib log messages, the wrong number of arguments were passed to `_log()`, resulting in the actual message not being logged. Additionally, since it did not pass in a filename that looked like a luakit module, `log_group_from_fct()` would often log a warning.

Since the filename and line number are often not available, and may not be as useful for log messages from an external library, I have removed them from the log, using "glib" as the log group. If this information is desired, I would suggest adding it to the log message (rather than the group) when available.


aidanholm:
Excellent! Thanks :)

PR #833, opened 2020-04-14
Avoid setting web process limit on WebKit 2.26

ulrikdem:
This is deprecated since WebKitGTK 2.26 and no longer has any effect.


aidanholm:
Thank you for the fix, and sorry for the delay in merging!

PR #837, opened 2020-04-18
gcc 10 build fix.

msdemlei:
This declares the three cross-module globals as extern in the headers
so the headers can safely be included in multiple modules.

The actual definitions are in luakit.c.

This would hopefully fix Debian bug https://bugs.debian.org/957511

### DIFF
--- a/clib/widget.h
+++ b/clib/widget.h
@@ -87,7 +87,7 @@ struct widget_t
     gpointer data;
 };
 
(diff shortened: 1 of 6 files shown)


aidanholm:
Thanks for the PR! Sorry for the terrible resolution time; I've squashed and merged this at 4efe52b8.

PR #839, opened 2020-04-27
Load domain-specific keys from persisted settings

### LINKED ISSUE #838: Domain-specific settings are not persisted correctly
# Bug Report

## Current Behavior

After setting a domain-specific setting key and restarting, that key is not available.

## Reproduction Instructions

1. Open luakit
2. `:seton example.com webview.zoom_level 300`
3. `:lua require'settings'.on["example.com"].webview.zoom_level`
  `=> 300` (ok)
4. restart luakit `:restart`
5. `:lua require'settings'.on["example.com"].webview.zoom_level`
  `=> nil` **(expected 300)**

## Expected Behavior

Domain settings should behave like global keys. I should be able to get the key I set, even after restart (like with global keys).
 
## Notes

The key is present in `$XDG_DATA_DIR/settings`. Maybe something goes wrong when loading the settings?

## Environment

linux distribution: Archlinux, kernel 5.6.6-arch1-1

```
$ luakit --version
luakit 2.1-33-g0de4b1c0
  built with webkit 2.28.1 (installed version: 2.28.2)
```

nicopap:
Domain-specfic settings are supported and serialized, and when loading
the settings file, were deserialized properly. But they never were added
to the settings tree.

Fixes #838

### DIFF
--- a/lib/settings.lua
+++ b/lib/settings.lua
@@ -31,7 +31,7 @@ do
     end)
     if not ok then persisted_settings = { domain = {}, } end
 
-    -- move .domain to .global[""]
+    -- move .global to .domain[""]
     local pgs = persisted_settings.domain[""] or {}
     persisted_settings.domain[""] = pgs
     if persisted_settings.global then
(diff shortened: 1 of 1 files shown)


aidanholm:
Thanks for the fix! The test suite is a little baroque, so i wouldn't worry too much about it. Unfortunately most of the lua code is only really testable through integration tests at the moment, and changing that would require a lot of cleanup.

PR #842, opened 2020-04-28
Fix re_match_text matching everything everytime

### LINKED ISSUE #841: Setting `follow.pattern_marker` to `follow.pattern_styles.re_match_text` doesn't work
### Current Behavior

Adding this to `userconf.lua`:
```lua
local follow = require("follow")
follow.pattern_maker = follow.pattern_styles.re_match_text
```

Breaks follow mode.

### Expected Behavior

When entering text in follow mode, it shoud select the proper links...

### Reproduction Instructions

* Set the lines I provided in `userconf.lua`
* open any page with links, press `f`
* type text in a link
* notice that nothing happens

### Environment

linux distribution: Archlinux, kernel 5.6.6-arch1-1
```
$ luakit --version
luakit 2.1-33-g0de4b1c0
  built with webkit 2.28.1 (installed version: 2.28.2)
```

nicopap:
Returing an empty string for the `hint` makes it so EVERY hint is
matched (since "" is a substring of every string). Returing `nil` (like
in `match_label`) fixes it. Fixes #841

### DIFF
--- a/lib/follow.lua
+++ b/lib/follow.lua
@@ -160,7 +160,7 @@ local function regex_escape(s)
     return s:gsub(escape_pat, "%%%1")
 end
 
-local re_match_text = function (text) return "", text end
+local re_match_text = function (text) return nil, text end
 local re_match_both = function (text) return text, text end
 local match_label_re_text = function (text)
     return #text > 0 and "^"..regex_escape(text) or "", text


aidanholm:
Thanks for the fix!

PR #844, opened 2020-05-08
Avoid a panic if there's problems reading /etc/hosts

okvik:
In case that, for whatever reason, the file /etc/hosts is missing
from the system or it otherwise can't be read an unprotected call
to io.lines() raises an error which bubbles up to panic the host.

This seems undesireable, certainly a full panic is not wanted.

Here we look the other way and pretend the hosts file was just empty.

### DIFF
--- a/lib/lousy/uri.lua
+++ b/lib/lousy/uri.lua
@@ -70,7 +70,8 @@ function _M.is_uri(s)
     -- Valid hostnames to check
     local hosts = {"localhost"}
     if get_setting("window.load_etc_hosts") then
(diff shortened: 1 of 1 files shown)


aidanholm:
Thanks for the fix! Please add a `local ok` line; this will inadvertently create a global variable called `ok`, which is causing the CI failure.

okvik:
Ah, that's quite a gotcha!

aidanholm:
merged; thanks again!

okvik:
Cheers!

PR #857, opened 2020-08-15
When a policy decision is requested, accept a (formal) status of 0 as…

### LINKED ISSUE #742: XHTML in a local file gives WebKitPolicyError-102
This is a bug report against luakit from both git HEAD (luakit
2.1-2-g9bbe4d69) and the last release (luakit 2.0.0).

## Current behaviour

If you create a file test.html with the content

```
<html><body><p>Hello world.</p></body></html>
```

and run

```
luakit -v file://$PWD/test.html
```

the expected web page appears.  Now, change test.html to

```
<html xmlns="http://www.w3.org/1999/xhtml">
<body><p>Hello world.</p></body></html>
```

```
luakit -v file://$PWD/test.html
```

will then fail with

```
[lua/error_page]: showing error page for error 'Frame load interrupted', code 'WebKitPolicyError-102', category 'ignore'
```

and then sit with ``about:blank`` in the address bar and an empty page.

## Expected behaviour

Well, it should show "Hello world" in the example.  

Frankly, I suspect the reason this fails is that something somewhere
looks for the string "&lt;html>" to detect when to render something into
HTML.  WebKitPolicyError-102, from reports to other webkit clients,
appears to be quite typically related to unexpected or undeclared media
types.  What I think should ideally happen is that luakit uses the OS's
media type detection (/etc/mime.types on unix) and then feed the
discovered media type to webkit.  

But then I give you that's perhaps overdoing it.  Just peeking for .html
in local file names would at least solve the problem reported here, I suspect.

## Environment

Linux Distribution & Version: Debian stretch with webkit from backports.

Output of `luakit --version`:

```
luakit 2.1-2-g9bbe4d69
  built with webkit 2.22.5 (installed version: 2.22.5)
```

msdemlei:
… success.

Background: webkit_uri_response_get_status_code(r) apparently returns 0
for local resources.  SOUP_STATUS_IS_SUCCESSFUL then is false, and a
fixed "no" policy is returned, leading to bug #742.

This fixes #742 for me, and it doesn't seem to affect anything else
adversely, though of course I can't really say if there are other conditions
when there is a 0 in response->status_code=0.

### DIFF
--- a/widgets/webview.c
+++ b/widgets/webview.c
@@ -526,7 +526,8 @@ decide_policy_cb(WebKitWebView* UNUSED(v), WebKitPolicyDecision *p,
         const gchar *uri = webkit_uri_response_get_uri(r);
         const gchar *mime = webkit_uri_response_get_mime_type(r);
 
-        if(!SOUP_STATUS_IS_SUCCESSFUL(webkit_uri_response_get_status_code(r)))
+        if(webkit_uri_response_get_status_code(r)
+            && !SOUP_STATUS_IS_SUCCESSFUL(webkit_uri_response_get_status_code(r)))
             return FALSE;
 
         luaH_object_push(L, w->ref);


aidanholm:
Good find; thanks for the fix!

PR #860, opened 2020-09-15
use os.remove() instead of luakit.spawn("rm %q")

semarie:
in session.lua, use os.remove() instead of luakit.spawn("rm %q")

### DIFF
--- a/lib/session.lua
+++ b/lib/session.lua
@@ -19,10 +19,6 @@ local _M = {}
 
 lousy.signal.setup(_M, true)
 
-local function rm(file)
-    luakit.spawn(string.format("rm %q", file))
-end
-
 --- Path to session file.
 -- @type string
 -- @readwrite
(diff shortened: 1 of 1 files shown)


semarie:
I wonder if there is any particular reason for having a spawn("rm %q") instead of simply use `os.remove()`.

I am looking to enable some sandbox facility using unveil(2) syscall on OpenBSD.

hippi777:
hi there! :) 

i believe it is for making things a bit faster by parallelization, but i think that probably spawning a process and running rm is slower, while removing files should be atomic, so only the number of files matter, but not the size of them... (i hope i didnt say anything dumb :D ) if not, then maybe whoever made it didnt think about the other way :D

bests! :)

c0dev0id:
The session state is kept in one file. And this is deleted on  `:q` and saved on `:wq`. I don't think this needs to be speed up by forking os commands.

I've tested this pull requests. Works as intended.

I'll merge it. Thank you!

PR #862, opened 2020-09-20
Don't destroy last view if close_with_last_tab=false

### LINKED ISSUE #725: luakit get stuck and start using all the CPU when closing the last tab

c0dev0id:
This is yet another attempt to fix https://github.com/luakit/luakit/issues/725. The problem here is that `w.tab:count()` is  1 when closing the second tab (1 left) but also on the last tab, because detach-tab does also detect this situation and does not detach it (so tabcount stays 1).

My version here properly detect this situation by saving the tab count before detach_tab is called and prevents to destroy the webview in this case.

As I don't have the issue on my machine, I'd like to ask @iamleot to retest.

### DIFF
--- a/common/ipc.c
+++ b/common/ipc.c
@@ -141,16 +141,17 @@ ipc_recv_and_dispatch_or_enqueue(ipc_endpoint_t *ipc)
         case G_IO_STATUS_EOF:
             verbose("g_io_channel_read_chars(): End Of File received");
             /* OSX and NetBSD are sending EOF on nonblocking channels first.
-             * This EOF can be ignored. They should end up in recv_hup(),
-             * but unfortunaltey they do not.
+             * These requests can be ignored. They should end up in
+             * recv_hup(), but unfortunately they do not.
              *
-             * In order to avoid an incref/decref loop. The refcount is
-             * forced down here until it reaches 1. Once 1 is reached,
-             * the decref call after this function will bring it to 0 and
-             * clean up the channel.
+             * If we do not close the socket, glib will continue to
+             * call the G_IO_IN handler.
+             *
+             * We decrement the refcount to 1 here, and when ipc_recv
+             * decrements the refcount to zero, the socket will be
+             * disconnected.
              */
-            if(&ipc->refcount > 1)
-                g_atomic_int_dec_and_test(&ipc->refcount);
+            g_atomic_int_dec_and_test(&ipc->refcount);
             return;
         case G_IO_STATUS_ERROR:
             if (!g_str_equal(ipc->name, "UI"))
--- a/common/ipc.c
+++ b/common/ipc.c
@@ -21,6 +21,7 @@
 #include "common/lualib.h"
 #include "common/luaserialize.h"
 #include "common/ipc.h"
+#include "log.h"
 
 /* Prototypes for ipc_recv_... functions */
 #define X(name) void ipc_recv_##name(ipc_endpoint_t *ipc, const void *msg, guint length);
@@ -138,6 +139,18 @@ ipc_recv_and_dispatch_or_enqueue(ipc_endpoint_t *ipc)
         case G_IO_STATUS_AGAIN:
             return;
         case G_IO_STATUS_EOF:
+            verbose("g_io_channel_read_chars(): End Of File received");
+            /* OSX and NetBSD are sending EOF on nonblocking channels first.
+             * This EOF can be ignored. They should end up in recv_hup(),
+             * but unfortunaltey they do not.
+             *
+             * In order to avoid an incref/decref loop. The refcount is
+             * forced down here until it reaches 1. Once 1 is reached,
+             * the decref call after this function will bring it to 0 and
+             * clean up the channel.
+             */
+            if(&ipc->refcount > 1)
+                g_atomic_int_dec_and_test(&ipc->refcount);
             return;
         case G_IO_STATUS_ERROR:
             if (!g_str_equal(ipc->name, "UI"))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Fixed bounding box not spanning over whole element.
 - Fixed an issue where styled hint labels caused intransparent bounding boxes.
+- Fixed a race condition when a tab is closed on NetBSD.
 
 ## [2.2]
 


c0dev0id:
Test do fail because:
```
RUN  async/test_memory_leaks / test_webview_from_closed_tab_is_released
  tabcount: 0
  w.tabs:count(): 1
FAIL async/test_memory_leaks / test_webview_from_closed_tab_is_released
  tests/async/test_memory_leaks.lua:18: webview widget not collected
  Traceback:
  (1) tests/async/test_memory_leaks.lua:18 in function [anonymous]
```

Maybe we can ignore this for the moment as this is the whole point of the diff.

iamleot:
I have just tested with current Git HEAD and I can confirm that with this patch luakit no longer gets stuck!

Thank you!

c0dev0id:
@iamleot: Thanks! However the tests were right, if you open up top you see that webkit processes accumulate and won't get destroyed. Not a great solution.

I noticed, that in the "last_tab" case, the detach function calls new_tab. And after that, destroy() is called. I'm thinking this might cause the trouble.

I've pushed another change that changes the order and calls new_tab later (after destroy()).

Would you mind testing this again?

c0dev0id:
Sorry for the force pushes. I tried to remove the config.mk file change, that snug in :( Not very successful though. Shouldn't matter for testing...

iamleot:
Don't worry about force pushes and/or unrelated commits!

With the updated patches - I have removed the hunks regarding `config.mk` - opening 3 tabs it seems it hangs also when trying to close the last tab.

So it seems problematic with that version.

c0dev0id:
Thanks. I tried another version where the last tab is not touched at all, but simply navigates to the newtab page. While this worked here, it prevented luakit from quitting as the can_quit() check entered an endless loop and consumed all CPU.

You're using luakit on NetBSD?

iamleot:
sdk writes:
> [...]
> You're using luakit on NetBSD?

Yes, all the tests I have done were with luakit on NetBSD.

c0dev0id:
Hi,
okay - I have NetBSD running on a spare lenovo x230 now.

*Notes:*
NetBSD does not have luajit, so I compiled with `USE_LUAJIT=0`
The freeze is happening here not only on the last tab, but on every attempt to close a tab.
A very relevant message here is `Error sending IPC message: Broken pipe`.

Another, probably unrelated issue: After installing, luakit did not find the config files in `/etc/xdg/luakit/`. I had to copy them over to `$HOME/.config/luakit/`.

@iamleot: If you're experiencing what I am experiencing here, then we have been looking into the wrong direction. Do you get the IPC Errors as well?

I'll play around a bit more.

c0dev0id:
Update: I found out that NetBSD has luajit. I didn't find the package because of the package name... "LuaJIT2".

I installed LuaJIT2, recompiled+reinstalled luakit. But all I get now when I start luakit is:
```
[    0.174922] E [core/common/util]: unprotected error in call to Lua API (runtime code generation failed, restricted kernel?)
Memory fault (core dumped)
```
@iamleot: can you help me out here? Are you using lua or luajit?

iamleot:
> Hi,
> okay - I have NetBSD running on a spare lenovo x230 now.
> 
> _Notes:_
> NetBSD does not have luajit, so I compiled with `USE_LUAJIT=0`
> The freeze is happening here not only on the last tab, but on every attempt to close a tab.
> A very relevant message here is `Error sending IPC message: Broken pipe`.
> 

At least in my tests I didn't see anything like that when calling `luakit -l DEBUG` (NetBSD/amd64 9.99.64, pkgsrc-current updated as yesterday with webkit-gtk-2.30.0 and `wip/luakit-git` built with `luajit` option (so using LuaJIT)).

> Another, probably unrelated issue: After installing, luakit did not find the config files in `/etc/xdg/luakit/`. I had to copy them over to `$HOME/.config/luakit/`.
>

I have used `pkgsrc/wip/luakit-git` and manually applied the patch of this PR to avoid such adjustments.

iamleot:
> Update: I found out that NetBSD has luajit. I didn't find the package because of the package name... "LuaJIT2".
> 
> I installed LuaJIT2, recompiled+reinstalled luakit. But all I get now when I start luakit is:
> 
> ```
> [    0.174922] E [core/common/util]: unprotected error in call to Lua API (runtime code generation failed, restricted kernel?)
> Memory fault (core dumped)
> ```
> 
> @iamleot: can you help me out here? Are you using lua or luajit?

I'm using LuaJIT and that's probably the JIT not happy with PaX MPROTECT.

Doing a `paxctl +m luakit` should address that.

c0dev0id:
Ok, paxctl did the trick. I have luakit running with luajit and I can reproduce the issue. I don't see the IPC error, but I see webkit processes starting with each new tab (expected). When I close one tab (no matter if it is the last one or not) luakit CPU usage spikes to 100%, luakit freezes, and the webkit process does *not* go away.

I think I can handle it from here.

c0dev0id:
I'm adding more information, but I don't know yet how to interprete it. Luakit doesn't actually freeze. The webview is frozen. When starting `luakit -v`, you can see that luakit in the background is reacting and you can properly quit it with ":q".

In dtruss, I saw that `_lwp_park()` is called and then a loop starts with `poll()` and `clock_gettime()`. This indicates that threads are waiting for a resource.

This is the dtruss summary recording a few seconds of the "freezing" state (main process):

```
CALL                                        COUNT
_lwp_park                                       3
clock_gettime                                 333
poll                                          333
recvmsg                                       333
write                                         665
read                                          666
```

The main process has 13 threads, which look like this:

```
Thread 13 (LWP 1 of process 1916):
#0  0x000072c8ac51b65a in recvmsg () from /usr/lib/libc.so.12
#1  0x000072c89d817d89 in _xcb_in_read () from /usr/X11R7/lib/libxcb.so.2
#2  0x000072c89d81873f in xcb_poll_for_event () from /usr/X11R7/lib/libxcb.so.2
#3  0x000072c8a24485c8 in ?? () from /usr/X11R7/lib/libX11.so.7
#4  0x000072c8a2448728 in ?? () from /usr/X11R7/lib/libX11.so.7
#5  0x000072c8a24489bb in _XEventsQueued () from /usr/X11R7/lib/libX11.so.7
#6  0x000072c8a242beac in XPending () from /usr/X11R7/lib/libX11.so.7
#7  0x000072c8b1a66a4b in gdk_event_source_prepare () from /usr/pkg/lib/libgdk-3.so.0
#8  0x000072c8b5848cc8 in g_main_context_prepare () from /usr/pkg/lib/libglib-2.0.so.0
#9  0x000072c8b5849580 in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#10 0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#11 0x000072c8b2015202 in gtk_main () from /usr/pkg/lib/libgtk-3.so.0
#12 0x000000000041646e in main (argc=1, argv=0x7f7fff652168) at luakit.c:254

Thread 12 (LWP 2 of process 1916):
#0  0x000072c8ac4a1f7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x000072c8aca0a71f in pthread_cond_timedwait () from /usr/lib/libpthread.so.1
#2  0x000072c8a00a1412 in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /usr/lib/libstdc++.so.9
#3  0x000072c8ae28da2e in bmalloc::Scavenger::threadRunLoop() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#4  0x000072c8ae28e0e1 in bmalloc::Scavenger::threadEntryPoint(bmalloc::Scavenger*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8a009f5aa in ?? () from /usr/lib/libstdc++.so.9
#6  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#7  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#8  0x00010102464c457f in ?? ()
#9  0x0000000000000000 in ?? ()

Thread 11 (LWP 3 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x000072c8b5bf37b0 in ?? ()
#10 0x000072c8b5bf37b0 in ?? ()
#11 0x0000000000000001 in ?? ()
#12 0x000072c852402678 in ?? ()
#13 0x000072c8524010d0 in ?? ()
#14 0x0000000000000000 in ?? ()

Thread 10 (LWP 4 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 9 (LWP 5 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 8 (LWP 6 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b5849740 in g_main_context_iteration () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8b5849776 in glib_worker_main () from /usr/pkg/lib/libglib-2.0.so.0
#5  0x000072c8b586e0c5 in g_thread_proxy () from /usr/pkg/lib/libglib-2.0.so.0
#6  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#7  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#8  0x0000000000000000 in ?? ()

Thread 7 (LWP 7 of process 1916):
#0  0x000072c8ac442c5a in accept () from /usr/lib/libc.so.12
#1  0x000072c8aca077e7 in accept () from /usr/lib/libpthread.so.1
#2  0x00000000004149f0 in web_extension_connect_thread (UNUSED_data=0x0) at ipc.c:161
#3  0x000072c8b586e0c5 in g_thread_proxy () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#5  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#6  0x00010102464c457f in ?? ()
#7  0x0000000000000000 in ?? ()

Thread 6 (LWP 8 of process 1916):
#0  0x000072c8ac4a1f7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x000072c8aca0a71f in pthread_cond_timedwait () from /usr/lib/libpthread.so.1
#2  0x000072c8b588d690 in g_cond_wait () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b581e6c8 in g_async_queue_pop_intern_unlocked () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8b586e61e in g_thread_pool_spawn_thread () from /usr/pkg/lib/libglib-2.0.so.0
#5  0x000072c8b586e0c5 in g_thread_proxy () from /usr/pkg/lib/libglib-2.0.so.0
#6  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#7  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#8  0x00010102464c457f in ?? ()
#9  0x0000000000000000 in ?? ()

Thread 5 (LWP 9 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8af6fc068 in gdbus_shared_thread_func () from /usr/pkg/lib/libgio-2.0.so.0
#5  0x000072c8b586e0c5 in g_thread_proxy () from /usr/pkg/lib/libglib-2.0.so.0
#6  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#7  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#8  0x0000000000000000 in ?? ()

Thread 4 (LWP 11 of process 1916):
#0  0x000072c8ac4a1f7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x000072c8aca0a71f in pthread_cond_timedwait () from /usr/lib/libpthread.so.1
#2  0x000072c8b588d690 in g_cond_wait () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b581e6c8 in g_async_queue_pop_intern_unlocked () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8b581ea32 in g_async_queue_pop () from /usr/pkg/lib/libglib-2.0.so.0
#5  0x000000000041b709 in ipc_send_thread (UNUSED_user_data=0x0) at common/ipc.c:75
#6  0x000072c8b586e0c5 in g_thread_proxy () from /usr/pkg/lib/libglib-2.0.so.0
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 3 (LWP 12 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 2 (LWP 13 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1

Thread 4 (LWP 11 of process 1916):
#0  0x000072c8ac4a1f7a in ___lwp_park60 () from /usr/lib/libc.so.12
#1  0x000072c8aca0a71f in pthread_cond_timedwait () from /usr/lib/libpthread.so.1
#2  0x000072c8b588d690 in g_cond_wait () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b581e6c8 in g_async_queue_pop_intern_unlocked () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8b581ea32 in g_async_queue_pop () from /usr/pkg/lib/libglib-2.0.so.0
#5  0x000000000041b709 in ipc_send_thread (UNUSED_user_data=0x0) at common/ipc.c:75
#6  0x000072c8b586e0c5 in g_thread_proxy () from /usr/pkg/lib/libglib-2.0.so.0
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 3 (LWP 12 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 2 (LWP 13 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()

Thread 1 (LWP 15 of process 1916):
#0  0x000072c8ac4429ba in poll () from /usr/lib/libc.so.12
#1  0x000072c8aca07d52 in poll () from /usr/lib/libpthread.so.1
#2  0x000072c8b584960a in g_main_context_iterate.isra () from /usr/pkg/lib/libglib-2.0.so.0
#3  0x000072c8b584991a in g_main_loop_run () from /usr/pkg/lib/libglib-2.0.so.0
#4  0x000072c8ae27fd3f in WTF::RunLoop::run() () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#5  0x000072c8ae257e1b in WTF::Thread::entryPoint(WTF::Thread::NewThreadContext*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#6  0x000072c8ae280bd0 in WTF::wtfThreadEntryPoint(void*) () from /usr/pkg/lib/libjavascriptcoregtk-4.0.so.18
#7  0x000072c8aca0c072 in ?? () from /usr/lib/libpthread.so.1
#8  0x000072c8ac487af0 in ?? () from /usr/lib/libc.so.12
#9  0x0000000000000000 in ?? ()
(gdb) 

```

c0dev0id:
I figured out what is looping.

In common/ipc.c the functions ipc_endpoint_incref and ipc_endpoint_decref are called in a endless loop.

I also found out that the issue is not the destruction of the old webview, but with creating the new one. The old WebViewWebProcess gets terminated successfully. The loop is starting when the new WebKitWebProcess is already running.

Sorry for the spam here... I'm really just starting to figure out this part of luakit.

c0dev0id:
In `common/ipc.c` function `ipc_recv_and_dispatch_or_enqueue()` the `g_io_channel_read_chars()` returns `G_IO_STATUS_EOF`. This is not happening on my other OpenBSD laptop. In fact, I've never seen EOF while browsing around. 

There are still 8 remaining bytes to read at this point but it hits `return`. So I assume this is where the loop is coming from. The bytes are never actually read so it's trying again and again and hitting EOF again and again.

header->done is 0. It's a bit strange to hit EOF at the end of the header...

Full loop with my added extra output:
```
[   34.124445] V [core/common/ipc]: IPC_ENDPOINT_INCREF START
[   34.124456] V [core/common/ipc]: OLD=1
[   34.124465] V [core/common/ipc]: IPC_ENDPOINT_INCREF RETURN: TRUE
[   34.124474] V [core/common/ipc]: ipc_recv_and_dispatch_or_enqueue start
[   34.124484] V [core/common/ipc]: read_chars: --- start ---
[   34.124494] V [core/common/ipc]: retcode: G_IO_STATUS_EOF
[   34.124736] V [core/common/ipc]: remaining bytes: 8
[   34.124747] V [core/common/ipc]: read bytes: 0
[   34.124756] V [core/common/ipc]: header->done: 0
[   34.124765] V [core/common/ipc]: hdr.length: 40
[   34.124770] V [core/common/ipc]: buf: '
[   34.124774] V [core/common/ipc]: read_chars: --- end ---
[   34.124783] V [core/common/ipc]: IPC_ENDPOINT_DECREF START
[   34.124793] V [core/common/ipc]: IPC_ENDPOINT_DECREF RETURN
[   34.124817] V [core/common/ipc]: IPC_ENDPOINT_INCREF START
[   34.124827] V [core/common/ipc]: OLD=1
[   34.124836] V [core/common/ipc]: IPC_ENDPOINT_INCREF RETURN: TRUE
[   34.124845] V [core/common/ipc]: ipc_recv_and_dispatch_or_enqueue start
[   34.124854] V [core/common/ipc]: read_chars: --- start ---
[   34.124865] V [core/common/ipc]: retcode: G_IO_STATUS_EOF
[   34.124874] V [core/common/ipc]: remaining bytes: 8
[   34.124883] V [core/common/ipc]: read bytes: 0
[   34.124892] V [core/common/ipc]: header->done: 0
[   34.124901] V [core/common/ipc]: hdr.length: 40
[   34.127125] V [core/common/ipc]: read_chars: --- end ---
[   34.127137] V [core/common/ipc]: IPC_ENDPOINT_DECREF START
[   34.127146] V [core/common/ipc]: IPC_ENDPOINT_DECREF RETURN
[   34.127170] V [core/common/ipc]: IPC_ENDPOINT_INCREF START
[   34.127194] V [core/common/ipc]: OLD=1
```
EDIT: Maybe relevant, the IPC type is eval_js.
```
Thread 1 "" hit Breakpoint 1, ipc_recv_and_dispatch_or_enqueue (ipc=0x7e8f61561e40) at common/ipc.c:157
157                 verbose("retcode: G_IO_STATUS_EOF");
1: *ipc = {name = 0x43c5f1 "UI", status = IPC_ENDPOINT_CONNECTED, channel = 0x7e8f62cd7b00, queue = 0x7e8f66f57e80, recv_state = {
    watch_in_id = 70, watch_hup_id = 71, queued_ipcs = 0x7e8f66f57ea0, hdr = {length = 39, type = IPC_TYPE_eval_js}, payload = 0x0, 
    bytes_read = 0, hdr_done = 0}, refcount = 2, creation_notified = 1}
(gdb) 
Continuing.
```

c0dev0id:
@aidanholm: do you have an idea here? `ipc_recv()` is called and `ipc_recv_and_dispatch_or_enqueue()` always returns with G_IO_STATUS_EOF. Nothing has been read (bytes_read = 0), 8 bytes are remaining.
I don't know what the error EOF indicates on a socket. It wants to read, but there is nothing?

`ipc_recv()` is registered as g_io channel watcher callback. This watcher calls ipc_recv repeatingly.

Is there a way to find the part of the code that triggers the watcher to respond? There is no `ipc_send()` call that I could relate to this.

I think the watcher sees 8 bytes and wants to have them read, but then EOF comes back, returns and that watcher tries again.

EDIT: found the calling place. it's the eval_js in javascript.c - It just comes in late and after the view has been destroyed.

c0dev0id:
Ah, this might be the issue. It could be that NetBSD also sends EOF first Looks like it. -> https://github.com/GNOME/glib/blob/10f2d10f1c9dc5e56d6a196389b8a8254262d8a0/glib/gtester.c#L223

c0dev0id:
Yes, it was the `eval_js` message. After disabling `eval_js` in `javascript.c`, the next message was the `scroll` message, which hit exactly the same problem.

So I think the solution is to ignore EOF.

c0dev0id:
I have an idea how to fix this.

c0dev0id:
@iamleot please retest. This works for me now.

iamleot:
@c0dev0id I can confirm that with your last patch the problem is fixed!

Thank you very much for all this analysis and for fixing it!

c0dev0id:
I just realized that I can just decrement the state refcount in the EOF handling. That brings the whole diff to one line.

I've added @aidanholm as reviewer because I'm not sure if something else must be done in addition before decrementing the refcount to not leak something. It might be okay as it is now, but I'd like to have an additional set of eyes on it.

iamleot:
I can confirm that also this latest version fixes #725.

aidanholm:
> Sorry for the spam here... I'm really just starting to figure out this part of luakit.

It's a little ugly, sorry... If i was writing this today it'd be much nicer... ¯\\\_(ツ)\_/¯

EOF means that the other side closed the socket. There's a separate handler for this that glib is supposed to call: `ipc_hup`. It does what you'd expect: decref the ipc object.

For `ipc_recv_and_dispatch_or_enqueue`, neither EOF or ERROR are handled: they just punt and return directly. This is fine: HUPs are supposed to be handled by `ipc_hup`. Note that ideally the call to `g_io_channel_read_chars` would first update the number of bytes read, before switching on the status: it's not entirely clear from the docs that you can't read some bytes and then return e.g. EAGAIN...

Anyway, the fact that you're seeing multiple calls to `ipc_recv` means that the `ipc_hup` isn't working for some reason... I think that's the root of the issue.

BTW: for decrementing the ipc refcount, use `ipc_endpoint_decref`: it does all the actual cleanup. `g_atomic_int_dec_and_test` just decrements the count :)

c0dev0id:
@aidanholm Ah, got it.

Remember, this is an `incref` `decref` loop. Refcount flips between 2 and 1. Incref -> 2, EOF, decref -> 1 and returns(!), then incref -> 2, EOF, decref -> 1, return... 

Decref does cleanup, but only when refcount reaches 0. This is never the case. This is the loop.

With this diff, I force-undo und prior incref. Then the decref afterwards counts to 0 and does clean up.

I think the diff here works fine in this situation. I will refine it a little and only decrement on EOF when refcount is > 1 to make sure that decref is called one last time to cleanup.

c0dev0id:
After `g_io_channel_read_chars` bytes_read is 0, remaining is 8. I'm not surprised by bytes_read, as reading anything would mean to read past the file-ending (whatever this means in a channel context). But there's nothing to read, thus EOF. This makes sense to me. I'm more confused about having 8 more bytes remaining.

I've seen a few implementations now that handle this special case for OSX and NetBSD. The condition is channel type `G_IO_IN` and `bytes_read=0` and status `G_IO_STATUS_EOF`. I'm not sure if checking for read_bytes is really necessary. I've never seen EOF ending up in G_IO_IN in any other case.

I think it's good as it is now.

common/ipc.c
```diff
@@ -138,6 +139,18 @@ ipc_recv_and_dispatch_or_enqueue(ipc_endpoint_t *ipc)
         case G_IO_STATUS_AGAIN:
             return;
         case G_IO_STATUS_EOF:
+            verbose("g_io_channel_read_chars(): End Of File received");
+            /* OSX and NetBSD are sending EOF on nonblocking channels first.
+             * This EOF can be ignored. They should end up in recv_hup(),
+             * but unfortunaltey they do not.
```
aidanholm: Typo: unfortunately

common/ipc.c
```diff
@@ -138,6 +139,18 @@ ipc_recv_and_dispatch_or_enqueue(ipc_endpoint_t *ipc)
...
+             *
+             * In order to avoid an incref/decref loop. The refcount is
+             * forced down here until it reaches 1. Once 1 is reached,
+             * the decref call after this function will bring it to 0 and
+             * clean up the channel.
+             */
+            if(&ipc->refcount > 1)
```
aidanholm: Add a space after `if`

aidanholm: But in general, I don't feel that this is really necessary. If we exclude the possibility that other parts of luakit decrement the refcount just before reaching here, then the only case it can be == 1 is if the 'channel open' ref is removed, when the refcount is decremented by `ipc_hup`. But this cannot happen: if `ipc_hup` was called, it would have been called and then returned before the `ipc_recv` handler was called by glib, and it would have already disconnected the ipc endpoint.

So this should really be an assertion: it's impossible for it to be <= 1, and if it *is* <= 1, it's a bug in luakit, not in glib or anything else.

common/ipc.c
```diff
@@ -138,6 +139,18 @@ ipc_recv_and_dispatch_or_enqueue(ipc_endpoint_t *ipc)
...
         case G_IO_STATUS_EOF:
+            verbose("g_io_channel_read_chars(): End Of File received");
+            /* OSX and NetBSD are sending EOF on nonblocking channels first.
+             * This EOF can be ignored. They should end up in recv_hup(),
+             * but unfortunaltey they do not.
+             *
+             * In order to avoid an incref/decref loop. The refcount is
+             * forced down here until it reaches 1. Once 1 is reached,
+             * the decref call after this function will bring it to 0 and
+             * clean up the channel.
```
aidanholm: I feel this should explain what is meant by the looping, in terms of the cause. I.e: "On OSX and NetBSD, if we do not close the socket, glib will continue to call the G_IO_IN handler."

The incref/decref behaviour is not relevant. That's just to ensure that the socket is alive while we read from it, but it's not related to this bug.

To be really picky, the refcount is not "forced down": that'd require a loop. I think it's enough to just say that "We decrement the refcount to 1 here, and when `ipc_recv` decrements the refcount to zero, the socket will be disconnected."

aidanholm (approved):
Just some nit picks, but overall this looks good!

Good work everyone on the diagnosis :)

common/ipc.c
c0dev0id: I'll remove the condition. This was more to safeguard my assumption.
EDIT: this also removes a warning because it was an  `gint * < int` comparison.

common/ipc.c
c0dev0id: You're right. I rephrased it with your suggestions. My throught was, that the refcount could be a lot higher and so every loop it would decrement. But in the EOF case, I haven't seen anything other than 2. But that was, where the comment came from (as well as the if-statement before). I adapted it.

iamleot:
Thank you very much @c0dev0id and @aidanholm!

PR #864, opened 2020-09-20
Fixing config.mk for cross building.

msdemlei:
This fixes Debian bug #923099, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=923099.

The package has a patch for this so far, but patching is always a chore.


aidanholm:
Thanks for the PR!

c0dev0id:
Tested on BSD as well. Thanks.

PR #887, opened 2021-01-23
Dark Mode Support

c0dev0id:
Hello,

I've implemented dark-mode support via command line toggle. Start luakit with `-d` and websites that support the css `prefer-color-scheme` attribute should be displayed in a dark fashion. You can try it on my [website](https://codevoid.de) - it should be dark with orange text with `-d`.

I personally don't mind having this as a command line toggle, but idealy it should be an entry in `:settings`. I tried, but was not able to accomplish reading a setting from the C layer.

The dark-mode is not a webkit attribute, but needs to be set on the gtk window. I think it can be set any time.

Anyone here that can help to bring this into `:settings`?

### DIFF
--- a/common/tokenize.list
+++ b/common/tokenize.list
@@ -121,6 +121,7 @@ selection
 select_region
 serif_font_family
 send_key
+set_dark_mode
 set_default_size
 set_title
 show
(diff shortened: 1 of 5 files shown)


c0dev0id:
I figured it out. Darkmode can now be activated under `:settings` -> `application.prefer_dark_mode`. It requires a restart of luakit to become active.

PR #893, opened 2021-01-30
Fixed injections and missing types in Gopher

Ygrex:
Added HTML encoding for injected text in:
- menu entries
- text files

Two missing types added I eventually came across somewhere in Gopherspace:
- CCSO (just readable representation in menu, no real support)
- PDF

### DIFF
--- a/lib/gopher.lua
+++ b/lib/gopher.lua
@@ -30,6 +30,7 @@ local function gophertype_to_text(gophertype)
         ["2"] = 'CCSO', -- CCSO Nameserver, not really supported type
         ["4"] = 'HEX ', -- BinHex-encoded
         ["5"] = 'DOS ', -- DOS file
+        ["6"] = 'UENC', -- uuencoded data
         ["7"] = 'FIND', -- search
         ["8"] = 'TEL ', -- telnet
         ["9"] = 'BIN ', -- binary
@@ -269,6 +270,8 @@ local function data_to_browser(data, url)
         mime = "application/mac-binhex40"
     elseif url.gophertype == "5" then
         mime = "application/octet-stream"
+    elseif url.gophertype == "6" then
+        mime = "text/x-uuencode"
     elseif url.gophertype == "9" then
         mime = "application/octet-stream"
     elseif url.gophertype == "d" then
--- a/lib/gopher.lua
+++ b/lib/gopher.lua
@@ -27,6 +27,7 @@ local function gophertype_to_text(gophertype)
     return ({
         ["0"] = 'TXT ', -- text file
         ["1"] = 'DIR ', -- submenu
+        ["2"] = 'CCSO', -- CCSO Nameserver, not really supported type
         ["4"] = 'HEX ', -- BinHex-encoded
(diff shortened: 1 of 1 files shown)


lib/gopher.lua
```diff
@@ -27,6 +27,7 @@ local function gophertype_to_text(gophertype)
     return ({
         ["0"] = 'TXT ', -- text file
         ["1"] = 'DIR ', -- submenu
+        ["2"] = 'CCSO', -- CCSO Nameserver, not really supported type
```
c0dev0id, line 30: While at it, we should probably add all canonical and "well known" types. Type 6 "UENC" would be missing.

lib/gopher.lua
```diff
@@ -38,6 +39,7 @@ local function gophertype_to_text(gophertype)
         ["I"] = 'IMG ', -- image
         ["M"] = 'MBOX', -- mbox
         ["p"] = 'IMG ', -- image
+        ["P"] = 'PDF ', -- binary pdf document
```
c0dev0id, line 43: I find "p" and "P" and "d" strange. I haven't see them yet as I would see "p" being "I" and "P" being type "9". Same for "M", which could also be "9".

Where are those types used?

I would see "P" as an exception because pdf can be handled by webkit. But most gopher clients wouldn't be able to handle it.

Gopherproxy can do: i0123456789ghs+IT:
https://git.codemadness.org/gopherproxy-c/file/gopherproxy.c.html

The same list is supported by sacc:
gopher://bitreich.org:70/1/scm/sacc/file/sacc.c.gph

The same list is supported by click:
gopher://bitreich.org:70/1/scm/clic/file/clic.lisp.gph

Bombardillo supports i0123456789gGhIpsST:
https://tildegit.org/sloum/bombadillo/src/branch/master/gopher/gopher.go

Lagrange supports i0134579gIhMs and defaults to application/octed-stream.
https://codeberg.org/skyjake/lagrange/src/branch/dev/src/gopher.c

Where have you seen "P" and which other client is supporting it?

c0dev0id (commented):
If you create a separate PR for the injections part, I'd submit that right away.

But I want to talk a little bit more about the gopher types.

Ygrex:
> If you create a separate PR for the injections part, I'd submit that right away.
> 
> But I want to talk a little bit more about the gopher types.

right, I thought about splitting PR, but these are just all changes I had on my local while Gopher was not yet accepted

lib/gopher.lua
Ygrex: I saw `P` selector occasionally, not sure I can find it right away, but quick veronica'ing gives some reason behind it:
gopher://whitemesa.net:70/0/gopher/gopher-types-from-UMN-sources.txt
```
#define A_PDF       'P'
```
no idea what the page is about, but at least somebody else saw `P` in use

lib/gopher.lua
Ygrex: good idea, added

c0dev0id:
Okay, I'll take it. Thank you!

Issue #879, opened 2021-01-02
System proxy/no proxy selection is not persisted


msdemlei:
## I'm submitting a…
Bug Report

## Current Behavior
In the proxy menu, you can select no or system proxy, and the choice
becomes active.  However, it is not persisted and thus once you have
defined a custom proxy, you cannot go back (within the proxy menu).

## Expected Behavior

The no proxy/system proxy settings should be persisted over sessions.

## Reproduction Instructions

Define a custom proxy.  Select it.

Then select proxy None or System.  Exit the browser, start a new
browser.  You'll see the custom proxy is still active.

## Environment

Linux Distribution & Version: Debian buster
Output of `luakit --version`:

```
luakit 2.0.0
  built with webkit 2.28.3 (installed version: 2.30.4)
```

## Considerations for a fix

Half of the trouble is that proxy.lua only saves custom proxies into
the datadir/proxymenu file, and the current selection is inferred from
where there's a star in there, as in

```
* sniff http://localhost:4080
  socks socks://localhost:5000
```

Now, the proxymenu Return binding doesn't call set_active for unnamed
rows (which both system and no proxy are) , hence this file isn't
written and the star persists.

A quick and dirty is is to insert a set_active(nil) as in:

```
          if row and row.address then
                if row.name then
                    _M.set_active(row.name)
                else
                    _M.set_active(nil)
                    soup.proxy_uri = row.address
                end
```

The reason I've not prepared a PR for this is that this doesn't tell
apart system proxy and no proxy.  From the API docs I can't tell how
this originally was intended to work.

Assuming there hasn't been a better plan: What I'd do is rewrite the menu
logic so no proxy and system proxy are treated like user-defined
proxies, i.e., receive a line in the proxymenu cache and aren't
artificially added internally.

c0dev0id:
I looked at the proxy module. I got the overall idea how it works, but I don't get why the design desicion was taken to implement it that way. I'd rather like to see the table to be like:
```
 |None|no_proxy
 |System|default
*|ActiveProxy|http://proxy:8080
 |InactiveProxy|socks://localhost:5555
```
It would always be possible to match the address for "default" and "no_proxy" and handle it accordingly. I played a little with it but didn't manage to get it to work the way I imagined. I might give it another spin. 

EDIT: It's not a good idea to change the delimiter format to stay backward compatible.

msdemlei:
On Thu, Jan 28, 2021 at 11:23:26AM -0800, sdk wrote:
[...]

PR #895, opened 2021-01-31
[WIP] Save last active proxy (even if it's none)

c0dev0id:
I was fiddling around with the proxymenu, so it persists the last use proxy, even when it is "None" or "System".

This seems to work, but only when the passmenu file is being deleted and recreated. This is not nice, but it's a start... 

This relates to https://github.com/luakit/luakit/issues/879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 - Removed debug symbol generation for default make.
 - Changed the C standard from gnu99 to c11 because Webkit wants it.
+- The proxy module saves when no proxy or system proxy was used last.
+- The proxy widget is hidden when proxy "None" is active.
 
 ### Fixed
 
--- a/lib/proxy.lua
+++ b/lib/proxy.lua
@@ -54,7 +54,7 @@ local active = noproxy
 local update_proxy_indicator = function (w)
     local name = _M.get_active().name
     local proxyi = w.sbar.r.proxyi
-    if name then
+    if name and name ~= "None" then
         local text = string.format("[%s]", name)
         if proxyi.text ~= text then proxyi.text = text end
         proxyi:show()
--- a/lib/proxy.lua
+++ b/lib/proxy.lua
@@ -181,19 +181,14 @@ window.add_signal("init", function (w)
     r.proxyi.font = theme.proxyi_sbar_font
     update_proxy_indicators()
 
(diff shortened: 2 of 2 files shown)


msdemlei:
Looks largely good to me – thanks a lot.

Two minor suggestions:

(a) When people already have a proxymenu file, they can't choose system
or no proxy any more when this is merged.  I'd hence say we should have
a changelog entry like

```
  If you have defined proxies, add:

    None no_proxy
    System default

  to your .local/share/luakit/proxymenu.  Otherwise, you may want
  to remove that file.
```

I'd have turned this into a PR except the CHANGELOG.md as it currently
stands doesn't really lend itself to this kind of prose.  Hm.  Is there
a better place for info like that?

Alternatively, we might want to add code that just adds these two entries
at the end of load if they're not there.  It's just two lines of code,
so perhaps it's a deal over a couple of lines of Changelog that nobody
reads...  But then it's code.  Hm.

(b) The [None] in the status bar when no proxy is selected might confuse
people, in particular because it's kind of the default.  I think the
original intention was to not show anything then, for which you'd need
to write

```
    if name~='None' then
```

around line 57.

c0dev0id:
I placed it here to hack on it. This should not be an incompatible change at all.

When the proxy entries are loaded, the plugin should check if the None/Default entries are there and if not, mix them in. On the next save they would be saved as well. I'll look at it again when I have some time.

c0dev0id:
@msdemlei Can you retest the latest change? For me it works in a backward compatible way.

msdemlei:
On Sat, Feb 13, 2021 at 12:45:52AM -0800, sdk wrote:
> @msdemlei Can you retest the latest change? For me it works in a
> backward compatible way.

Works just great for me, thanks.

I was trying to figure out how unconditionally adding the None and
System proxies might hurt but could not find a credible one (there's
no way anyone could want to unconfigure one of them, and the
"active" indicator can't be on them if they're not already
present...)

So, I'd say this is good for merging.

c0dev0id:
I occationally see double None/System entries in the menu. It seems to happen when I start luakit with a URL on the command line. When I start luakit and then navigate somewhere, it behaves correct.

aidanholm:
This looks good to me -- it'd be nice to debug duplicate none/system entries, but I think saving the proxy setting over restarts is more important

Issue #889, opened 2021-01-25
[FEATURE REQUEST] swich to one of the already openned tabs


ngortheone:
Hi, I was not able to find how to acheive this with Luakit:

Description:
Les say a user has 5 tabs open.
It would be great to have a command that would allow swiching to openned tabs with autocomplete.
Similar to how `:tabopen <TAB>` shows history autocompletion show autocompletion of exiting tabs. Switch to one of the tabs based on user's input

Is it possible do in Luakit?

c0dev0id:
You can do this already :)

try typing `<tabnumber> gt`, so `5gt` jumps to tab 5. 

![changetab](https://user-images.githubusercontent.com/44190565/105738098-f9473c00-5f36-11eb-86c9-8f2f07afce4f.png)

EDIT: if you switch favicons off, the tabs have numbers instead.

taobert:
There's not (to my knowledge) an inbuilt tab-completey method.
It could be done (and i think it would be useful), but you should probably not hold your breath on someone doing it for you.

As @c0dev0id points out, you can use `gt` to switch tabs if you know the index of the tab you want.
This would work quite well in the "5 tabs open" use case you suggested.
Edit: Further, `Alt-1` to `Alt-0` will switch to tab 1 to 10.

If you struggle to remember the indices of every tab you have open,
(i currently have 48 tabs in my "Luakit" tabgroup, and hide the tablist,) a couple of alternatives would be:
`tabmenu` in the `luakit-plugins` repository will give you a menu of tabs that you can scroll through and select from.
Or, you could spawn `dmenu` which will  do the same, and also let you type in a pattern to restrict the tabs shown.
A reference implementation of the latter might look like:
```
modes.add_binds("normal", {
    { "<control-t>", "List tabs in dmenu",
	function (w)
	    local list = ""
	    local dmenu='dmenu -i -l 20'
	    local label

	    for i, view in ipairs(w.tabs.children) do
		label=( (view.title ~= "Web Process Crashed" and view.title)
			or view.uri
			or "" )
		list = list .. i ..  "  " ..  label .. "\n"
	    end
	    list = lousy.util.escape(list)

	    local fh=io.popen("echo '" ..list .. "' |" .. dmenu ,"r")
	    local sel=fh:read("*a")
	local tabchoice = string.match(sel, "^[0-9]+")
	w.tabs:switch(tabchoice)
	end },
})

```

c0dev0id:
[...]

PR #914, opened 2021-03-25
Add tabgroups module from `luakit-plugins` to `luakit`

taobert:
This moves the `tabgroups` module from `luakit-plugins` into the main repo (per opinions expressed in #889).
It also adds a `tgname` widget to display current tab group name on status bar.

Changes made to the tabgroups module to make it non-broken are:
- Correct sorting
- Deal with nil title
- Emit `switched-page` on tabgroup name change (to trigger redrawing)
- Avoid crashing when restoring session with empty tabgroup
- Meet style requirements for `make run-tests`

and in addition:
- Add a `current_tabgroup()` function to export the name of the current tabgroup (used by `tgname`)

For ease of review, these changes (to the existing tabgroups plugin) are available as a draft PR against `luakit-plugins` ([#29](https://github.com/luakit/luakit-plugins/pull/29))

Note that `tabgroups` emits `switched-page` on tabgroup switching (and with this PR on tabgroup renaming).
I can instead add a `tabgroup-changed` signal if that's prefered.
Note also that i've added a commented out `require "tabgroups"` to rc.lua.
My preference, if it were agreeable, would be to have it uncommented.

### DIFF
--- a/config/rc.lua
+++ b/config/rc.lua
@@ -106,6 +106,10 @@ local tabhistory = require "tabhistory"
 -- Add command to list open tabs
 local tabmenu = require "tabmenu"
 
+-- Allow for tabs to be grouped together.
+-- One tab group is displayed in a window at any given time.
+--local tabgroups = require "tabgroups"
+
 -- Add gopher protocol support (this module needs luasocket)
 -- local gopher = require "gopher"
 
--- a/lib/lousy/widget/init.lua
+++ b/lib/lousy/widget/init.lua
@@ -18,6 +18,7 @@ local wrapped = {
     tabi     = wrap("lousy.widget.tabi"),
     tablist  = wrap("lousy.widget.tablist"),
     tab      = wrap("lousy.widget.tab"),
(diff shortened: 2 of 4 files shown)


config/rc.lua
```diff
@@ -106,6 +106,9 @@ local tabhistory = require "tabhistory"
 -- Add command to list open tabs
 local tabmenu = require "tabmenu"
 
+-- Add command to list open tabs
```
c0dev0id: Copy paste from command above? ;-)

taobert: Indubitably. Thanks. Corrected now.

taobert:
I just tweaked the `restore` signal handler to have a single `local group = w2groups[w].groups[gn]` definition (there had been two in slightly different scopes).
Edit: No, that made things bad, so i've undone it. Sorry for the noise.

c0dev0id:
I'm not using the tabgroup heavily, but it also didn't disturb my browsing or cause noticable side effects. So I think it can go in. Thanks!

PR #915, opened 2021-03-25
Remove compile warnings

taobert:
This corrects some compile time warnings that i'm seeing. It:
- Removes a deprecation warning about `gdk_keymap_get_default()`
- Changes a `lua_pushinteger` call to `lua_pushlightuserdata` to remove an "integer from pointer" complaint.
- Defines a `PD_CASE` macro (by analogy with PI_CASE) and replaces a PI_CASE call to remove an "integer from pointer" complaint.

Although these changes _feel_ right to me and the browser still seems to function,
i don't grok the C-LUA interface well enough to trust my judgement,
and i've not gone to the effort of making sure that those lines have been executed in my testing.

### DIFF
--- a/common/util.h
+++ b/common/util.h
@@ -47,11 +47,12 @@
 #define WARN_UNUSED __attribute__ ((warn_unused_result))
 
 /* stack pushing macros */
-#define PB_CASE(t, b) case L_TK_##t: lua_pushboolean   (L, b); return 1;
-#define PF_CASE(t, f) case L_TK_##t: lua_pushcfunction (L, f); return 1;
-#define PI_CASE(t, i) case L_TK_##t: lua_pushinteger   (L, i); return 1;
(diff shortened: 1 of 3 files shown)


c0dev0id:
Oh nice! I had the first two on my list to fix. I tried and failed and didn't try again.
I'll test this over the next days. On a first glance, it looks fine to me.

taobert:
RE `PD_CASE`, i chose "D" for data. In retrospect "P" for pointer may have been a better choice.
I don't think it's all that important either way, but if there were a percieved benefit to "P",
it'd be better to change it at this stage.

taobert:
I just corrected
"#define PD_CASE(t, **i**) case L_TK_##t: lua_pushlightuserdata (L, d); return 1;"
to
"#define PD_CASE(t, **d**) case L_TK_##t: lua_pushlightuserdata (L, d); return 1;"

c0dev0id:
Thanks!

PR #920, opened 2021-04-16
Correct/remove centre-finding behaviour in `click` evaluator (closes #916)

### LINKED ISSUE #916: Luakit not being able to open hinted links
**Current Behavior:**
I am not able to open some of the hinted links generated by pressing 'f' key.
It doesn't happen all the time, but it is somewhat annoying considering that for every bugged pages, I would have to reload/restart luakit, or use ';o' to temporarily fix the issue.  

Below is a video clip of the issue.
https://user-images.githubusercontent.com/72597973/112709696-876b7e80-8ef6-11eb-939d-b2aebb44bd69.mp4
I was constantly pressing 'r', but nothing seems to be executing whatsoever.

**Desired Behavior:**
Have luakit open up links to be followed.

**How can we reproduce it (step by step):**
I have no idea, is it only me who has been affected?

**Environment:**

Linux Distribution & Version:
Output of `luakit --version`:
luakit 2.0.0
  built with webkit 2.30.1 (installed version: 2.30.6)

**Note about webkit issues:**

If you're reporting a rendering issue, please test it with the gnome
browser ephiphany as well. If the issue occurs there too, we're very
likely not able to help. These issues should be reported to webkit:
https://bugs.webkit.org

taobert:
So the discussion about how to best close #916 rather petered out.

This PR intentionally has two separate commits to more easily facilitate the removal of the more drastic (second) change.
The first (6c09e81b) subtracts the viewport from the `element.rect`, correcting the current behaviour.
The second (2045a50) deletes this clause entirely, so the hinted element is clicked, rather than it's central child.
My belief (#916 for rationale) is that this clause is unnecessary,
but there _may_ be valid reasons for it being there which i've not considered,
in which case reverting 2045a50 will still leave the situation corrected.

### DIFF
--- a/lib/follow_wm.lua
+++ b/lib/follow_wm.lua
@@ -30,7 +30,10 @@ local evaluators = {
         if element.child_count > 0 then
             local r = element.rect
             local doc = element.owner_document
-            element = doc:element_from_point(r.left + r.width/2, r.top + r.height/2) or element
+            local scroll = page:eval_js([=[ window.scrollX + ' ' + window.scrollY; ]=])
+            local scrollX, scrollY = scroll:match("^(%S+) (%S+)$")
+            element = doc:element_from_point(r.left - scrollX + r.width/2,
+                                             r.top - scrollY + r.height/2)
         end
         element:click()
     end,


c0dev0id:
Hi,

I agree with you. The original code came in here: https://github.com/luakit/luakit/commit/ffd43c4f8ccb04087dd809b789936f9b413f8279 and then was refactored to follow_wm.lua later on.

I don't see any issue that lead to this code to be added in the commit history. So I assume it was a (reasonable) implementation desicion to click links in the center. A user would likely click there too.

I pulled this change into my daily luakit. If it doesn't act up in the next days, we can pull it in and see if anyone else has an issue with it.

Thanks for looking into this!

taobert:
Ok, i'll flinch first.

The selectors used in `follow.lua` are css selectors, and don't look at JS clickability,
so my rationale about clickable child elements having their own hints doesn't hold up.
Although i don't have examples, i think it's reasonably likely that ablating the clause will cause something somewhere to stop working.
My current grasp of current JS usage is inadequate to know for sure,
but it seems like a plausible web monkey hack to wrap a JS enabled image in `<a>` tags to provide the expected change in mouse cursor, and present a friendly sounding link wherever a browser shows that.
Although the obvious use of this (to me) would be obnoxious uri rewriting, it probably has more benign applications.
So, at this stage, i'd urge that only the first of the commits be added to the main repo.
Do you need me to edit the PR?

c0dev0id:
> Do you need me to edit the PR?

That would actually be nice. I don't know how to change + merge a PR.

taobert:
Ok, this branch now subtracts the current viewport and clicks the centre element.

c0dev0id:
Thanks!

PR #933, opened 2021-06-05
Update nav buttons to allow paging

### LINKED ISSUE #932: Bookmarks limited to 100
**Current Behavior:**
If the `bookmarks.db` has more than 100 bookmarks, `:bookmarks` does not show them all.

**Desired Behavior:**
A sane value or unlimited bookmarks should be shown when a user opens `:bookmarks`

**How can we reproduce it (step by step):**
1. Open `luakit`
2. Add over 100 bookmarks (or add them using `sqlite3` then open `luakit`)
3. Type `:bookmarks` and only 100 will show up

Adding `1000` to the following line allowed the amount of bookmarks I have to show. I set it to `65536` for now, but having a limit may not be necessary:

https://github.com/luakit/luakit/blob/ad2ce7e2fb80049a960989d3c9a840c837879ae4/lib/bookmarks_chrome.lua#L286

**Environment:**

Linux Distribution & Version: Manjaro (up-to-date)
Output of `luakit --version`: 2.3

taobert:
Call `updateNavButtons` to close #932.

### DIFF
--- a/lib/bookmarks_chrome.lua
+++ b/lib/bookmarks_chrome.lua
@@ -343,6 +343,7 @@ window.addEventListener('load', () => {
             }
 
             $results.innerHTML = results.map(makeBookmark).join('')
+            updateNavButtons()
         })
     }
 


ghost:
Just an FYI, the cursor type on hover for the `Next` and `Prev` buttons is a text entry cursor. Otherwise this one-liner fixes the issue with the `const limit` value reverted to `100`.

taobert:
I hadn't noticed the text cursor.
This is because the "Buttons" are `a` elements which don't have any `href` attribute (but presumably have some JS that activates them). I can see that this is unintuitive, and adding `href=#` is sufficient to correct the behaviour, however it's consistent with the history chrome page (and possibly others) so i'll do a bit more digging and treat it as a separate issue.

c0dev0id:
Thanks!

PR #935, opened 2021-07-01
Updated Makefile so LIBDIR actually does something

Charadon:
While packaging luakit for opensuse, I found it was a struggle to get the library to it's proper location (lib64) and after investigating, found that the LIBDIR variable is unused. So I went and fixed it so LIBDIR can actually do something.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,8 +104,8 @@ install: all
 	install -m644 lib/*.* $(DESTDIR)$(PREFIX)/share/luakit/lib
 	install -m644 lib/lousy/*.* $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy
 	install -m644 lib/lousy/widget/*.* $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
-	install -d $(DESTDIR)$(LIBDIR)/luakit
-	install -m644 luakit.so $(DESTDIR)$(LIBDIR)/luakit/luakit.so
+	install -d $(DESTDIR)$(LIBDIR)
+	install -m644 luakit.so $(DESTDIR)$(LIBDIR)/luakit.so
 	install -d $(DESTDIR)$(PREFIX)/bin
 	install luakit $(DESTDIR)$(PREFIX)/bin/luakit
 	install -d $(DESTDIR)$(XDGPREFIX)/luakit/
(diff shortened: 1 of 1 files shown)


taobert (commented):
I may be off the mark, but as the default `LIBDIR` is "LIBDIR     ?= $(PREFIX)/lib/luakit", i would expect that these should have the "/luakit" removed from them.

Charadon:
> I may be off the mark, but as the default `LIBDIR` is "LIBDIR ?= $(PREFIX)/lib/luakit", i would expect that these should have the "/luakit" removed from them.

If you're refering to config.mk, I didn't even think about changing that, but it shouldn't be hard change. Want me to go ahead and change it?

taobert:
No, i'm sorry i wasn't clear.
Currently, `LIBDIR` is set in `config.mk` to `$(PREFIX)/lib/luakit`, but then (as you point out) is not used.
Instead, the value `$(PREFIX)/lib/luakit` is used directly, so currently, `luakit.so` is put in `$(DESTDIR)$(PREFIX)/lib/luakit/`.
Your change now installs to `$(DESTDIR)$(LIBDIR)/luakit`, which when `LIBDIR` is expanded becomes `$(DESTDIR)$(PREFIX)/lib/luakit/luakit/`, so there's an extra `luakit/` directory created.
I feel that you should instead create and install to `$(DESTDIR)$(LIBDIR)`, which would maintain the current behaviour (with the current defaults), and also allow people to set `LIBDIR` to omit the `luakit` subdirectory entirely if they so choose.

Charadon:
Okay, I updated the Makefile to do that

taobert:
Looking at the makefile, there's an `options:` clause, that should probably have LIBDIR added to it (since it now does something).

I should also mention that my opinions aren't necessarily requisite for PR inclusion, and that this will have to wait until a grown-up casts an eye over it before it's pulled.

Charadon:
Alright, added a LIBDIR echo under options. Now I guess we'll wait a see if this gets approved.

Makefile
```diff
@@ -32,6 +32,7 @@ options:
 	@echo "XDGPREFIX    = $(XDGPREFIX)"
 	@echo "PIXMAPDIR    = $(PIXMAPDIR)"
 	@echo "APPDIR       = $(APPDIR)"
+	@echo "APPDIR       = $(LIBDIR)"
```
c0dev0id: Copy paste error. This should be LIBDIR on the left side.

c0dev0id (changes requested):

c0dev0id:
Looks good. Thank you!

PR #941, opened 2021-07-18
Substitute full commit hash in getversion.sh

arcnmx:
Short hashes do not generate reproducible archives; [downloading an archive](https://github.com/luakit/luakit/archive/cb591ba9a466140559a5d9fcd5652b8aea13d80c.tar.gz) a couple times can produce at least 2 versions containing different short hash lengths, which [breaks tools and distros that rely on downloaded archive integrity hashes being stable](https://github.com/NixOS/nixpkgs/issues/84312). Presumably this is because github servers use different git versions/defaults/whatever to produce the archive, or maybe because as the repo history grows a given shorthash length may become ambiguous.

Always substituting the full commit hash (and then truncating it in the script to a fixed short length) addresses the problem.

### DIFF
--- a/build-utils/getversion.sh
+++ b/build-utils/getversion.sh
@@ -9,8 +9,8 @@
 #   ./build-utils/getversion.sh export-subst
 #
 # This tells git to replace the format string in the following line with the
-# current short hash upon the calling of the `git archive <hash/tag>` command.
-VERSION_FROM_ARCHIVE=$Format:%h$
+# current hash upon the calling of the `git archive <hash/tag>` command.
+VERSION_FROM_ARCHIVE=$Format:%H$
 
 # The preferred method is to use the git describe command but this is only
 # possible if the .git directory is present.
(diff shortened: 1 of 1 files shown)


c0dev0id:
Distributions and packages are encouraged to use the releases and not download generated tag or commit versions. Commit hashes are not the only problem. If you rely on on-the-fly generated archives, the checksum is not stable either. Releases are stable in terms of commit hashes and checksums.

But I'm okay with this change.

arcnmx:
> Distributions and packages are encouraged to...

FWIW I had encountered this just trying to make a luakit-develop/luakit-git style package, and nixpkgs happens to prefer to substitute tarballs for git clones by default.

> ... use the releases and not download generated tag or commit versions.

I may be blind but I'm not able to find any official release tarballs by looking through the website or github uploads. Github's releases by default seem to just link to the generated tarballs by tag ref, which have the same problem and do not produce reliable archives.

> Commit hashes are not the only problem. If you rely on on-the-fly generated archives, the checksum is not stable either.

Well a build tool can choose to sanitize the tarball or extract it to work around this, but export-subst is unique in that it changes the actual contents of the source archive being downloaded, which can be surprising!

> But I'm okay with this change.

Cool, it's always good to be consistent and explicit anyway!

PR #949, opened 2021-08-18
Add SVG desktop icon support

trepetti:
The `pixmap` directory in `share` can also use SVG files for HiDPI/scalable desktop icons, so this is an easy fix.


c0dev0id:
I'm not using a WM that makes use of these icons. But my question would be: Do we still ne the png version then?

trepetti:
I think the PNG is still good to have as a fallback. Several other programs I know provide both in case the DE doesn't support SVG icons. Thank you for the review!

PR #976, opened 2021-12-16
Document luakit.spawn(_sync) behavior

### LINKED ISSUE #973: Safety considerations for luakit.spawn
`luakit.spawn` is documented as "parsed with a simple shell-like parser".   Are there considerations I need to think about to avoid shell injection?  For example from the 02-faq.md:

    if string.match(string.lower(uri), "^mailto:") then
        luakit.spawn(string.format("%s %q", "urxvt -title mutt -e mutt", uri))
        return false
    end

Is there a reason this kind of string formatting is used?  Do I need to avoid certain use cases?

Would there be any interest in a spawn function that took a list of arguments and didn't use any kind of shell or shell-like parsing?

samuelallan72:
This is for the `luakit.spawn` and `luakit.spawn_sync` commands.

It would be nice to document this so users can see how their spawn commands will be interpreted.

Closes #973

### DIFF
--- a/doc/luadoc/luakit.lua
+++ b/doc/luadoc/luakit.lua
@@ -71,14 +71,14 @@
 -- @tparam integer status The exit status code of the process. Its meaning is system-dependent.
 
 --- Spawn a process asynchronously.
--- @tparam string cmd The command to execute. It is parsed with a simple shell-like parser.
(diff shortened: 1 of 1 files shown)


c0dev0id:
Thank you ;)

PR #979, opened 2022-01-06
Solaris fixes for build/install

pekdon:
Small fixes to get luakit build and install on Solaris 10.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,14 +94,15 @@ clean:
 	rm -rf doc/apidocs doc/html luakit $(OBJS) $(EXT_OBJS) $(TSRC) $(THEAD) buildopts.h luakit.1 luakit.1.gz luakit.so
 
 install: all
-	install -d $(DESTDIR)$(PREFIX)/share/luakit/
-	install -d $(DESTDIR)$(DOCDIR) $(DESTDIR)$(DOCDIR)/classes $(DESTDIR)$(DOCDIR)/modules $(DESTDIR)$(DOCDIR)/pages
+	install -d $(DESTDIR)$(DOCDIR)/classes
+	install -d $(DESTDIR)$(DOCDIR)/modules
+	install -d $(DESTDIR)$(DOCDIR)/pages
 	install -m644 README.md AUTHORS COPYING.GPLv3 $(DESTDIR)$(DOCDIR)
 	install -m644 doc/apidocs/classes/* $(DESTDIR)$(DOCDIR)/classes
 	install -m644 doc/apidocs/modules/* $(DESTDIR)$(DOCDIR)/modules
 	install -m644 doc/apidocs/pages/* $(DESTDIR)$(DOCDIR)/pages
 	install -m644 doc/apidocs/*.html $(DESTDIR)$(DOCDIR)
-	install -d $(DESTDIR)$(PREFIX)/share/luakit/lib $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
+	install -d $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
 	install -m644 lib/*.* $(DESTDIR)$(PREFIX)/share/luakit/lib
 	install -m644 lib/lousy/*.* $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy
 	install -m644 lib/lousy/widget/*.* $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
--- a/Makefile
+++ b/Makefile
@@ -118,9 +118,8 @@ install: all
 	install -m644 extras/luakit.desktop $(DESTDIR)$(APPDIR)
 	install -d $(DESTDIR)$(MANPREFIX)/man1/
 	install -m644 luakit.1.gz $(DESTDIR)$(MANPREFIX)/man1/
-	mkdir -p resources
-	find resources -type d -exec install -d $(DESTDIR)$(PREFIX)/share/luakit/'{}' \;
-	find resources -type f -exec sh -c 'f="{}"; install -m644 "$$f" "$(DESTDIR)$(PREFIX)/share/luakit/$$(dirname $$f)"' \;
+	install -d $(DESTDIR)$(PREFIX)/share/luakit/resources/icons
+	for i in resources/icons/*; do install -m644 "$$i" "$(DESTDIR)$(PREFIX)/share/luakit/resources/icons"; done
 
 uninstall:
 	rm -rf $(DESTDIR)$(PREFIX)/bin/luakit $(DESTDIR)$(PREFIX)/share/luakit $(DESTDIR)$(PREFIX)/lib/luakit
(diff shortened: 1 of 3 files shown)


c0dev0id (commented):
comment moved to review.

Makefile
```diff
@@ -95,13 +95,18 @@ clean:
...
+	install -d $(DESTDIR)$(DOCDIR)/pages
 	install -m644 README.md AUTHORS COPYING.GPLv3 $(DESTDIR)$(DOCDIR)
 	install -m644 doc/apidocs/classes/* $(DESTDIR)$(DOCDIR)/classes
 	install -m644 doc/apidocs/modules/* $(DESTDIR)$(DOCDIR)/modules
 	install -m644 doc/apidocs/pages/* $(DESTDIR)$(DOCDIR)/pages
 	install -m644 doc/apidocs/*.html $(DESTDIR)$(DOCDIR)
-	install -d $(DESTDIR)$(PREFIX)/share/luakit/lib $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
+	install -d $(DESTDIR)$(PREFIX)/share/luakit/lib
+	install -d $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy
+	install -d $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
```
c0dev0id, line 105: This is actually not necessary. `install -d` creates missing components. So only the longest path needs to be given.

pekdon:
That would just be expanded by the shell and passed to install as before so won't work unfortunately. :(

build-utils/getversion.sh
```diff
@@ -16,7 +16,7 @@ VERSION_FROM_ARCHIVE=$Format:%H$
 # possible if the .git directory is present.
 if [ -d .git -a -r .git ]
 then
-    VERSION_FROM_GIT=$(git describe --tags --always)
+    VERSION_FROM_GIT=`git describe --tags --always`
```
c0dev0id, line 19: I'm not a fan - but I know the stone age sh in solaris demands it.
Does one of these work?
```
VERSION_FROM_GIT != "git describe --tags --always"
VERSION_FROM_GIT = $(shell git describe --tags --always)
```
I'd prefer !=

Makefile
```diff
@@ -95,13 +95,18 @@ clean:
 
 install: all
 	install -d $(DESTDIR)$(PREFIX)/share/luakit/
-	install -d $(DESTDIR)$(DOCDIR) $(DESTDIR)$(DOCDIR)/classes $(DESTDIR)$(DOCDIR)/modules $(DESTDIR)$(DOCDIR)/pages
+	install -d $(DESTDIR)$(DOCDIR)
+	install -d $(DESTDIR)$(DOCDIR)/classes
+	install -d $(DESTDIR)$(DOCDIR)/modules
+	install -d $(DESTDIR)$(DOCDIR)/pages
```
c0dev0id, line 99: Can you try if this works on solaris?
`install -d $(DESTDIR)$(DOCDIR)/{classes,modules,pages}`

c0dev0id (changes requested):

Makefile
c0dev0id: Ok, due to shell expansion. Of course.

Makefile
c0dev0id: As per your comment earlier - this would not work. I'm ok with this.

Makefile
pekdon: Yeah, I noticed that while doing the change but kept it as is. I'll drop the extra install.

build-utils/getversion.sh
pekdon: That's Makefile syntax, is it not? And no, does not work, not even with bash.

Makefile
```diff
@@ -94,14 +94,15 @@ clean:
 	rm -rf doc/apidocs doc/html luakit $(OBJS) $(EXT_OBJS) $(TSRC) $(THEAD) buildopts.h luakit.1 luakit.1.gz luakit.so
 
 install: all
-	install -d $(DESTDIR)$(PREFIX)/share/luakit/
```
taobert, line 97: I'm cautious about deleting this, as it may diminish futureproofness and hackability.
As has been mentioned in other comments, `install -d` will make any required parent directories,
so the following `install -d $(DESTDIR)$(DOCDIR)/classes` will create `$(DESTDIR)$(PREFIX)/share/luakit/`,
but only because `DOCDIR` is _by default_ `$(PREFIX)/share/luakit/doc`.
If a user decides for whatever reason that they want docs elsewhere, the rest of the install should carry on as normal.

taobert: Hrmm. Well, looking harder, even if `DOCDIR` changes, `$(DESTDIR)$(PREFIX)/share/luakit/` will be created before it's needed by `install -d $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget` below. So sorry for the noise i guess.

taobert (commented):
While i have no particular objection to for loops, i'm curious whether the failure of the current find(s) is due solely to solaris' sh requiring backticks rather than `$(...)`. Does
```
find resources -type f -exec sh -c 'f="{}"; install -m644 "$$f" "$(DESTDIR)$(PREFIX)/share/luakit/`dirname $$f`"' \;
 ```
work?

**Comment by pekdon** [comment]
> While i have no particular objection to for loops, i'm curious whether the failure of the current find(s) is due solely to solaris' sh requiring backticks rather than `$(...)`. Does
> 
> ```
> find resources -type f -exec sh -c 'f="{}"; install -m644 "$$f" "$(DESTDIR)$(PREFIX)/share/luakit/`dirname $$f`"' \;
> ```
> 
> work?

Unfortunately not, I'm actually not sure how to trick find into parsing {} there as what happens is that f gets set to {} literal and not the current path.

```
# works
-exec echo '{}' \;
# prints {}
-exec echo "'{}'" \;
# prints {}
-exec echo '"{}"' \;
```

File: Makefile (hunk 4 above)
Comment (pekdon) [reply Makefile:97]: Want me to re-add the "extra" install -d or keep them? As long as you reviewers agree I'll update to whatever you decide. :)

File: build-utils/getversion.sh (hunk 2 above)
Comment (c0dev0id) [reply build-utils/getversion.sh:19]: Interesting. Both versions work here with gmake. != also with bsd make. Backticks are fine. Let's leave them in.

File: Makefile (hunk 4 above)
Comment (taobert) [reply Makefile:97]: Well i'm not someone you need to convince, i'm just sticking my nose in unnecessarily.

My comment above was (erroniously) trying to protect against breakage for users who redefine variables. 
I _do_ kinda feel that explicitly creating every directory which will be installed into is likely to reduce brittleness,
and it's my supposition that that's why we have (had) lines like 
```
install -d $(DESTDIR)$(PREFIX)/share/luakit/lib $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy $(DESTDIR)$(PREFIX)/share/luakit/lib/lousy/widget
```
rather than just using the longest path. So if it were up to me (it's not), i'd leave them in. If there were good reasons for reducing the number of `install` invocations (i don't really see any) then i'd put in a comment drawing the readers attention to the fact that the implicitly created directories are required. In a perfect world nobody working on the code would need those cues, but we don't have one of those, and they don't do much harm. But, i don't see these changes actively breaking anything right now, so, ya-know, whatever.:)

c0dev0id (approved):

c0dev0id:
I'm merging this PR. I think it's fine functionality wise. We can still improve things afterwards.
Thanks @pekdon

PR #987, opened 2022-03-16
Search LUAKIT_CONFIG_PATH for rc.lua and theme.lua

pekdon:
When building with a non-standard XDGPREFIX luakit fails to start as it
is unable to find rc.lua and then later theme.lua if the user lacks
these files in the configured path.

Add the configured LUAKIT_CONFIG_PATH to the list of paths to look in
for a smoother experience with custom installs.

### DIFF
--- a/clib/luakit.c
+++ b/clib/luakit.c
@@ -739,6 +739,7 @@ luaH_luakit_index(lua_State *L)
       /* push string properties */
       PS_CASE(CACHE_DIR,        globalconf.cache_dir)
       PS_CASE(CONFIG_DIR,       globalconf.config_dir)
+      PS_CASE(XDGPREFIX,        LUAKIT_CONFIG_PATH)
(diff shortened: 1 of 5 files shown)


pekdon:
Anyone interested in taking a look at this? Something you want, or am I just doing it plain wrong?

aidanholm:
I'm sorry for the slow response -- It's been a while since I worked on luakit, but as far as I can tell [luakit.config_dir here](https://github.com/luakit/luakit/pull/987/files#diff-dafdb4cd9c65630e0eb4f663f962d59ae114be208db067f01caf79ed388a511eR335) should be set from XDGPREFIX, as it's initialized at build time [here](https://github.com/luakit/luakit/blob/b3be1f8549464b7d48f832a74b8388385a024a81/Makefile). Could you help check whether `:lua luakit.config_dir` prints the expected path to stdout?

PR #1001, opened 2022-07-29
Fix typo in rc.lua

txgk:


config/rc.lua
```diff
@@ -110,7 +110,7 @@ local tabmenu = require "tabmenu"
 -- One tab group is displayed in a window at any given time.
 --local tabgroups = require "tabgroups"
 
--- Add gopher protocol support (this module needs luasocket)
+-- Add gopher protocol support (this module needs lua-socket)
```
c0dev0id: It's luasocket. See https://lunarmodules.github.io/luasocket/

The package name may be different on different OS.

txgk: Yeah, right. Sorry I thought package name in my distro is truth in the first instance :D

c0dev0id:
Thanks ;)

PR #1003, opened 2022-08-26
Use foreground colour for cursor (closes #1002)

### LINKED ISSUE #1002: Change cursor color in input bar
I've looked through previous issues, grepped the code, looked at the documentation, and searched the web, but there seems to be no recorded way of changing cursor color in the input bar. In my case, because of my system's theme, it is too close to the background color.

Despite there not being a way within luakit to customize the cursor shape (as per #859), I'd suppose there is one to customize color, as all other colors around it (text foreground and background, titles of menus, etc) are customizable.

Is there a way to do it?

taobert:
Use foreground colour for cursor (closes #1002)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fixed build on Solaris.
 - Fixed relocation errors on Sparc64.
 - Fixed proxymenu to show default entries when proxymenu file is not present.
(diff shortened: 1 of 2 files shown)


taobert:
As i mention in #1002, making the cursor independently theme-able may be preferred,
however this will at least ensure readabillity in the meantime.

PR #1018, opened 2022-10-05
Mention FAQ, Quick Start, etc on :help

taobert:
The default luakit homepage (`luakit.github.io`) has a `Documentation` link which takes users to the API Documentation page.
This page has (in addition to API Documentation) links to a number of pages which are of use to
new and prospective users, including the Quick Start Guide and the FAQ.
These pages, although useful, are arguably not API Documentation, and should probably be filed elsewhere.

By contrast, if a user misses the Documentation link at the top of the home page, but has the nous to type in `:help`,
they get useful information on which `luakit://` chrome pages are available, with Settings and Binds pointed out explicitly,
but they will only see the Quick Start Guide, FAQ etc if they navigate the API Index link.
A new user who would benefit from these pages, is unlikely to click a link to the API Documentation,
and so is likely to not see these pages.

This PR duplicates the links from the top of the API Doc page to the top of the `:help` page.

### DIFF
--- a/lib/help_chrome.lua
+++ b/lib/help_chrome.lua
@@ -41,6 +41,18 @@ local index_html_template = [==[
             license.  It is primarily targeted at power users, developers and any people with too much time
             on their hands who want to have fine-grained control over their web browser&rsquo;s behaviour and
             interface.</p>
+            <p> 
+            Useful (though outdated) documentation can be found here:
+            <ul>
+            <li><a href="doc/pages/01-authors.html">Authors</a></li>
+            <li><a href="doc/pages/02-faq.html">Frequently Asked Questions</a></li>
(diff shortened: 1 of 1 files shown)


taobert:
Sorry, i'm not sure how i managed to incorporate so many commits into this PR. 
It should have only been `a346301`.
I'll try to pare it down.

taobert:
Ok. Better now.

msdemlei:
While I can't say I'm too happy about the "though outdated", I'd say this is a reasonable thing -- you have my vote.

But perhaps we can start an effort to bring the docs up to date -- where do you see particularly bad obsolescence?

c0dev0id:
Well, this is a start ;) Thank you!

Issue #1030, opened 2023-01-01
Password fields are hinted based on their hidden/masked content


matteopt:
**Current Behavior:**
When pressing `f` to enter follow mode, password fields are hinted based on their (hidden/masked) content.

**Desired Behavior:**
Never hint password fields based on their hidden/masked content.

**How can we reproduce it (step by step):**
1. Open any login page (e.g. Google)
2. Enter something in the password field (it may help to enter something that's already on the page, e.g. "Forgot")
3. Enter follow mode and search for what's been entered in the password field
4. The password field will be hinted or followed/focused

**Hint to fix this:**
Around [L286](https://github.com/luakit/luakit/blob/develop/lib/select_wm.lua#L286) of select_wm.lua  
I'm not sure what properties `element` has but it should be simple to check if input type equals password

**Environment:**

Linux Distribution & Version: Arch Linux
Output of `luakit --version`:
```
luakit 2.3.3
  built with webkit 2.38.2 (installed version: 2.38.3)
```

**Note about webkit issues:**

If you're reporting a rendering issue, please test it with the gnome
browser ephiphany as well. If the issue occurs there too, we're very
likely not able to help. These issues should be reported to webkit:
https://bugs.webkit.org

taobert:
Thanks for noticing.
I can't reproduce this, but i believe you.
Would you try this patch:
```
diff --git a/lib/select_wm.lua b/lib/select_wm.lua
index 94dc53ee..3af01420 100644
--- a/lib/select_wm.lua
+++ b/lib/select_wm.lua
@@ -283,9 +283,12 @@ local function frame_find_hints(page, frame, elements)
         local rbb = get_element_bb_if_visible(element,wbb, page)
 
         if rbb then
-            local text = element.text_content
-            if text == "" then text = element.value or "" end
-            if text == "" then text = element.attr.placeholder or "" end
+            local text = ""
+            if element.type ~= "password" then
+                local text = element.text_content
+                if text == "" then text = element.value or "" end
+                if text == "" then text = element.attr.placeholder or "" end
+            end
             hints[#hints+1] = { elem = element, bb = rbb, text = text }
         end
     end
```

In this context `element` is a `dom_element`, and it's properties are documented in the API available [online](http://luakit.github.io/docs/classes/dom_element.html), and as a luakit chrome page: luakit://help/doc/classes/dom_element.html.


You could also avoid the behaviour you're describing by only matching against the label text:
```lua
local follow = require "follow
[...]

PR #1033, opened 2023-01-04
Disabled matching masked password field content when hinting

matteopt:
As discussed in https://github.com/luakit/luakit/issues/1030, password fields, i.e. `input` elements with `type="password"`, were being matched based on their masked content during hint generation.  
This PR changes the matching behaviour to ignore `text_content` and `value` for any element with `type="password"`.


c0dev0id:
It took me a while to understand it, but I can reproduce it. Therefore I merged the PR. Thanks!

Issue #1025, opened 2022-12-05
Migrate to webkit2gtk-4.1, javascriptcore-4.1 and libsoup-3.0


fictitiousexistence:
Updated gentoo system today including webkit-gtk to 2.38.2-r500

Tried to rebuild luakit master and got the following message:

`Package webkit2gtk-4.0 was not found in the pkg-config search path.`
`Perhaps you should add the directory containing webkit2gtk-4.0.pc'`
`to the PKG_CONFIG_PATH environment variable`
`Package 'webkit2gtk-4.0', required by 'virtual:world', not found`
`Package 'javascriptcoregtk-4.0', required by 'virtual:world', not found`
`config.mk:109: *** Cannot find required package(s\) to build luakit. Please check` `you have the above packages installed and try again.  Stop.`

Then I found this [https://discourse.gnome.org/t/webkit2gtk-4-1-added-to-gnome-sdk-webkit2gtk-4-0-deprecated/7053](https://discourse.gnome.org/t/webkit2gtk-4-1-added-to-gnome-sdk-webkit2gtk-4-0-deprecated/7053)

Thanks

c0dev0id:
Hi,

Most distributions are allowing to install both at the same time.

I briefly tried to switch luakit to 4.1 by changing webkit2gtk-4.0 and javascriptcoregtk-4.0 to 4.1 in `config.mk`. It will then also pull in libsoup-3.0, which is incompatible.

I've created branch [webkit2gtk-4.1](https://github.com/luakit/luakit/tree/webkit2gtk-4.1) to work on. I'm not sure yet how much effort this is. A minimum would be to work through the libsoup migration document here: https://libsoup.org/libsoup-3.0/migrating-from-libsoup-2.html

taobert:
I've [cloned](https://github.com/taobert/luakit/tree/webkit2gtk-4.1) that branch, and updated the soup bits so that it compiles and runs well enough to type this in.
The code's too messy to PR it to `develop`, and i don't trust that the changes haven't screwed up memory management, 
but it might be useful as a reference.

fictitiousexistence:
Thanks @taobert 
I just ran it for the last hour for my normal morning routines and did not run into any issues. 
I usually open up 20+ tabs.

c0dev0id:
Cool! I was also working on it, but mine is not compiling yet. I'm pulling your change in and we continue from there. You can always send PRs towards this branch. I'll merge them when they work for me.

[...]

PR #1034, opened 2023-01-14
Webkit2gtk 4.1 (#1025)

taobert:
Ok, so here's the next iteration towards closing #1025.
In my hands, this builds and runs against both webkit2gtk-4.1 & 4.0,
so it should be pullable into develop, pending code review and testing.
It's only been tested on Debian Sid.

### common/clib/soup.h:
- This uses `g_uri_join_with_user()` to build the URI string.
  This is somewhat more elegant, and avoids creating a GUri object.
- It no longer produces URIs with passwords, to retain consistency with previous behaviour.
- It dances around the g_uri_* functions using -1 in place of a missing port.
- It adds a soup version check which includes the appropriate header,
  and #defines a bitmap that's absent in soup2.

I'm not sure why `common/clib/soup.h`'s functions are declared static, or are in a header file at all.
They had previously been in `clib/soup.c`,
but were moved to `common/clib/soup.h` in a283b1b99 ("Split soup library into UI, Web, and shared parts").

### common/property.c
- Removed the comments with the rationales for the changes.
- Duplicate the soup version test from `common/clib/soup.h`.

I'm pretty sure that the `URI:` cases in `common/property.c` are never triggered, as a `property_value_t` of `URI` never gets used.
This makes it more difficult to confirm that these clauses are functioning correctly.

### luakit.c:
- Added some extra information to the `--version` command line output.
It was useful for me to figure out what library version i needed to test,
and might be of value in the future, but is certainly not required for SOUP3 compilation.

### widgets/webview/auth.c
- Remove #includes for system soup headers.
One no longer exists with soup3, the other two don't seem necessary for compilation.
They'd been there since 2011, and i suspect have become redundant ("soup" doesn't appear in the file).
However, i haven't actually tested http authentication with them deleted.

### CHANGELOG.md
- Note the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Develop
+
+### Fixed
+
+- Update uri handling for SOUP3
+
 ## [2.3.3]
 
 ### Fixed
--- a/common/clib/soup.h
+++ b/common/clib/soup.h
@@ -21,6 +21,19 @@
 #ifndef LUAKIT_COMMON_CLIB_SOUP_H
 #define LUAKIT_COMMON_CLIB_SOUP_H
 
+#include <libsoup/soup-version.h>
+#if SOUP_CHECK_VERSION(3,0,0)
+#include <libsoup/soup-uri-utils.h>
+#else
+#include <libsoup/soup-uri.h>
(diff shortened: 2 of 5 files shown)


c0dev0id:
You're really doing all the work here! Good to see, that you found a better solution for the GET_PROP parameter duplication.

PR #1035, opened 2023-01-17
Correct indentation in :lua command

taobert:
The indentaton in the `:lua` command's implementation gets a bit skewed.
This corrects it.


aidanholm:
thanks!

PR #1036, opened 2023-01-24
Support for Webkit2gtk 4.1 and libsoup 3

c0dev0id:
@taobert worked on the Webkit2gtk 4.1 support, which requires luakit to migrate to libsoup 3.0.
Issue: https://github.com/luakit/luakit/issues/1025

The migration steps are mostly documented here:
https://libsoup.org/libsoup-3.0/migrating-from-libsoup-2.html

This PR is to gather feedback, clean up the code a bit and then hopefully merge it relatively soon.

### DIFF
--- a/clib/request.c
+++ b/clib/request.c
@@ -85,7 +85,7 @@ luaH_request_finish(lua_State *L)
     size_t length;
     const gchar *data = lua_tolstring(L, 2, &length);
     const gchar *mime = lua_tostring(L, 3) ?: "text/html";
-    GInputStream *stream = g_memory_input_stream_new_from_data(g_memdup(data, length), length, g_free);
+    GInputStream *stream = g_memory_input_stream_new_from_data(g_memdup2(data, length), length, g_free);
     webkit_uri_scheme_request_finish(request->request, stream, length, mime);
     g_object_unref(stream);
 
--- a/clib/unique.c
+++ b/clib/unique.c
@@ -77,8 +77,13 @@ luaH_unique_new(lua_State *L)
     }
 
     GError *error = NULL;
-    if (!globalconf.application)
+    if (!globalconf.application) {
+#if GTK_CHECK_VERSION(2,74,0)
         globalconf.application = gtk_application_new(name, G_APPLICATION_DEFAULT_FLAGS);
+#else
+        globalconf.application = gtk_application_new(name, G_APPLICATION_FLAGS_NONE);
+#endif
+    }
 
     g_application_register(G_APPLICATION(globalconf.application), NULL, &error);
     if (error != NULL) {
--- a/clib/unique.c
+++ b/clib/unique.c
@@ -78,7 +78,7 @@ luaH_unique_new(lua_State *L)
 
     GError *error = NULL;
     if (!globalconf.application)
-        globalconf.application = gtk_application_new(name, G_APPLICATION_FLAGS_NONE);
+        globalconf.application = gtk_application_new(name, G_APPLICATION_DEFAULT_FLAGS);
 
     g_application_register(G_APPLICATION(globalconf.application), NULL, &error);
     if (error != NULL) {
--- a/common/clib/soup.h
+++ b/common/clib/soup.h
@@ -21,6 +21,19 @@
 #ifndef LUAKIT_COMMON_CLIB_SOUP_H
 #define LUAKIT_COMMON_CLIB_SOUP_H
 
+#include <libsoup/soup-version.h>
+#if SOUP_CHECK_VERSION(3,0,0)
+#include <libsoup/soup-uri-utils.h>
+#else
+#include <libsoup/soup-uri.h>
+#define SOUP_HTTP_URI_FLAGS (G_URI_FLAGS_HAS_PASSWORD     |\
+                             G_URI_FLAGS_ENCODED_PATH     |\
+                             G_URI_FLAGS_ENCODED_QUERY    |\
+                             G_URI_FLAGS_ENCODED_FRAGMENT |\
+                             G_URI_FLAGS_SCHEME_NORMALIZE)
+#endif
+
+
 static GRegex *scheme_reg;
 
 static gint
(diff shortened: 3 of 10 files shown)


clib/request.c
```diff
@@ -85,7 +85,7 @@ luaH_request_finish(lua_State *L)
     size_t length;
     const gchar *data = lua_tolstring(L, 2, &length);
     const gchar *mime = lua_tostring(L, 3) ?: "text/html";
-    GInputStream *stream = g_memory_input_stream_new_from_data(g_memdup(data, length), length, g_free);
+    GInputStream *stream = g_memory_input_stream_new_from_data(g_memdup2(data, length), length, g_free);
```
c0dev0id, line 88: OK

Unrelated to the migration. (deprecated function)

clib/unique.c
```diff
@@ -78,7 +78,7 @@ luaH_unique_new(lua_State *L)
 
     GError *error = NULL;
     if (!globalconf.application)
-        globalconf.application = gtk_application_new(name, G_APPLICATION_FLAGS_NONE);
+        globalconf.application = gtk_application_new(name, G_APPLICATION_DEFAULT_FLAGS);
```
c0dev0id, line 82: OK.

FLAGS_NONE didn't follow the naming scheme, so it was renamed and has the same semantics.

common/clib/soup.h
```diff
@@ -30,15 +43,19 @@ luaH_soup_uri_tostring(lua_State *L)
     gint port;
     /* check for uri table */
     luaH_checktable(L, 1);
-    /* create empty soup uri object */
-    SoupURI *uri = soup_uri_new(NULL);
-    soup_uri_set_scheme(uri, "http");
```
c0dev0id, line 35: Have we lost that the scheme is set to http by default?

common/property.c
```diff
@@ -22,13 +22,24 @@
 #include <lauxlib.h>
 #include "common/property.h"
 
-#include <libsoup/soup-uri.h>
 #include <webkit2/webkit2.h>
+#include <libsoup/soup-version.h>
+#if SOUP_CHECK_VERSION(3,0,0)
```
c0dev0id, line 27: Would luakit still compile with libsoup2? I tried to deinstall libsoup3, but it would deinstall half my system. I'm almost certain we can drop this check.

common/property.c
```diff
@@ -51,9 +62,9 @@ luaH_gobject_get(lua_State *L, property_t *p, GObject *object)
 
       case URI:
         g_object_get(object, p->name, &u, NULL);
-        tmp.c = u ? soup_uri_to_string(u, 0) : NULL;
+        tmp.c = u ? g_uri_to_string_partial (u, G_URI_HIDE_PASSWORD) : NULL;
         lua_pushstring(L, tmp.c);
-        if (u) soup_uri_free(u);
+        if (u) g_uri_unref(u);
```
c0dev0id, line 67: OK.

The `g_uri_to_string_partial` part has the same semantic according the the migration document and `g_uri_unref`works. `g_free` does too, but GUri documents that `g_uri_unref` should be used.

common/property.c
```diff
@@ -103,13 +114,17 @@ luaH_gobject_set(lua_State *L, property_t *p, gint vidx, GObject *object)
...
         else
             tmp.c = g_strdup_printf("http://%s", tmp.c);
-        u = soup_uri_new(tmp.c);
-        gboolean valid = !u || SOUP_URI_VALID_FOR_HTTP(u);
+        u = g_uri_parse(tmp.c, SOUP_HTTP_URI_FLAGS, NULL);
+
+        gboolean valid = !u || ( (!g_strcmp0(g_uri_get_scheme(u), "http")
+                                     || !g_strcmp0(g_uri_get_scheme(u), "https") )
+                                 && g_uri_get_host(u)
+                                 && g_uri_get_path(u) );
```
c0dev0id, line 122: OK.

common/clib/soup.h
```diff
@@ -47,42 +64,54 @@ luaH_soup_uri_tostring(lua_State *L)
...
+                                NULL,  // omit password to retain soup_uri_to_string()'s behaviour
+                                NULL,  // auth_params
+                                host,
+                                port,
+                                path,
+                                query,
+                                fragment);
+
+    lua_pushstring(L, uri);
+    g_free(uri);
```
c0dev0id, line 97: `g_uri_unref`?

common/clib/soup.h
```diff
@@ -93,9 +122,10 @@ luaH_soup_push_uri(lua_State *L, SoupURI *uri)
     PUSH_PROP(query)
     PUSH_PROP(fragment)
 
-    if (uri->port) {
+    port = g_uri_get_port(uri);
+    if (port > 0) {
         lua_pushliteral(L, "port");
-        lua_pushnumber(L, uri->port);
+        lua_pushnumber(L, port);
```
c0dev0id, line 128: OK.

common/clib/soup.h
```diff
@@ -118,13 +148,14 @@ luaH_soup_parse_uri(lua_State *L)
...
+    GUri *uri = g_uri_parse (str, SOUP_HTTP_URI_FLAGS, NULL);
     g_free(str);
     if (uri) {
         luaH_soup_push_uri(L, uri);
-        soup_uri_free(uri);
+        g_uri_unref(uri);
+        return 1;
     }
-    return uri ? 1 : 0;
+    return 0;
```
c0dev0id, line 158: OK.

common/clib/soup.h
```diff
@@ -47,42 +64,54 @@ luaH_soup_uri_tostring(lua_State *L)
...
-    if (soup_uri_get_scheme(uri) == SOUP_URI_SCHEME_FILE) {
-        soup_uri_set_host(uri, "");
+    if (!g_strcmp0(scheme, "file")) {
+        /* I assume that this is strdup()ed by g_uri_join(), so use
+         * some space on the stack rather than calloc'ing */
+        host = "";
     }
 
     GET_PROP(user)
-    GET_PROP(password)
```
c0dev0id, line 55: I'm still struggling to understand what's going on here.

c0dev0id (commented):
Looks pretty good already and all unit tests pass. I have to reread the issue again to undestand what's going on with the password GET_PROP. But it works.

The --version output looks a bit excessive now. But that's fine for me if the numbers will help to solve problems.

common/clib/soup.h
taobert: No, `uri` here is a `gchar *` not a `GUri`.

common/property.c
taobert: It compiles (and the test is required) with:
```
PKGS += webkit2gtk-4.0
...
PKGS += javascriptcoregtk-4.0
```
in `config.mk`

I haven't uninstalled soup3 to double check that it's not somehow being pulled in.

If we want to jettison soup2 support and force an upgrade to `...-4.1` then the test isn't needed.
I feel that it's worthwhile retaining soup2 support until nobody wants it.

common/clib/soup.h
taobert: I'm not sure which bit you're refering to, but that doesn't mean you're wrong.

The soup2 behaviour is that a `SoupURI` is a struct containing `char *`s to hold the various components of a uri (and and int for `port`). `soup_uri_set_*` are setter functions for those componenets. These setter functions `strdup` and `free` their respective componenets internaly, however we don't need this data to be saved past the end of the function, so (i think) this shouldn't be necessary.
The new code instead has explicit `gchar*`s to which are assigned the components.
These componenets are returned from Lua-space, and i believe Lua will deal with any required garbage collection, so we shouldn't need to free these. I may be wrong about this. Also, possibly the `lua_pop` removes the object from Lua-space, and allows for garbage collection, but that in my testing, it just hadn't been collected _yet_. If this is the case, then we'll need to `strdup` before the `lua_pop` and then free all the `gchar*`s at the end. I think it's probably _not_ the case, because Lua should be able to call `uri_tostring` without it's URI being freed, but perhaps an extra copy takes place at the Lua-C interface. My grasp of the Lua-C interface is inadequate here.


`password` is omited to retain the previous behaviour:
["Note that the output will never contain a password, even if uri does."](https://libsoup.org/libsoup-2.4/SoupURI.html#soup-uri-to-string).
This is tested for by `make run-tests`.

common/clib/soup.h
taobert: Yes, and that's an error. This should be changed to:
```
const gchar * scheme   = "http";
```

clib/unique.c
aidanholm: Can/should this use an ifdef to retain compatibility with gio < 2.74 ?

common/property.c
```diff
@@ -103,13 +114,17 @@ luaH_gobject_set(lua_State *L, property_t *p, gint vidx, GObject *object)
...
         if (valid) {
             g_object_set(object, p->name, u, NULL);
             g_free(tmp.c);
         }
-        soup_uri_free(u);
+        g_uri_unref(u);
```
aidanholm: Does this work as expected if `u` is NULL?

aidanholm (approved):

common/property.c
c0dev0id: I'm having trouble getting to this code path. I placed `exit(5);` in front of `g_uri_unref` and ran the test suite - no issues. Normal operations with the default config.. browsing, opening tabs/windows, settings page etc. are working.

common/property.c
c0dev0id: It works.  You can verify with `ldd` which libraries are pulled in. I thought GUri would be new in libsoup 3 - but it's not. So this version check is indeed enough to keep it working with libsoup 2.

common/clib/soup.h
c0dev0id: Oh, my question was about the last line, because I didn't see why we got rid of `GET_PROP(password)` here. Got it now. Thanks.

common/property.c
taobert: As i mentioned in the original pull request (#1034):
>I'm pretty sure that the URI: cases in common/property.c are never triggered,
as a property_value_t of URI never gets used.
This makes it more difficult to confirm that these clauses are functioning correctly.

@aidanholm's question is valid, but needs to be confirmed with a MWE or at least a reference to the relevant docs.

taobert: ...`soup_uri_free` and `g_uri_unref` both call `g_return_if_fail (uri != NULL);` as their first line, so the behaviour isn't changed.
If `G_DISABLE_CHECKS` is false,
`g_return_if_fail` in turn calls `return_if_fail_warning()` which is commented as: "/* Treat this like an assert(0). */".
If `G_DISABLE_CHECKS` is true, `g_return_if_fail` is no-opped away:
> #define g_return_if_fail(expr) G_STMT_START{ (void)0; }G_STMT_END

but has the somewhat omminous comment that:
>...
If @expr evaluates to %FALSE, the current function should be considered to
have undefined behaviour (a programmer error).
...

So it would be better if these were _not_ called with NULL.
`luaH_gobject_get` tested for NULL before calling `soup_uri_free`.
`luaH_gobject_set` doesn't, but probably should.

Issue #782, opened 2019-04-30
I can't setup custom characters for hints


delister:
I try setup custom characters for hint. In my rc.lua I added code from example:
```
 -- Add vimperator-like link hinting & following
local follow = require "follow"
-- Use a custom charater set for hint labels
local s = follow.label_styles
follow.label_maker = s.sort(s.reverse(s.charset("asdfqwerzxcv"))) 
```

But when I run luakit, I get error:

`[core/common/lualib]: Lua error: attempt to index local s (a nil value)`

I checked documentation: https://luakit.github.io/docs/modules/follow.html and didn't see property `label_styles`. And here I found another example:
```
local select = require "select"

select.label_maker = function ()
    local chars = charset("asdfqwerzxcv")
    return trim(sort(reverse(chars)))
end
```

But I got error:
` [core/common/lualib]: Lua error: attempt to call global charset (a nil value)`

How I can to setup it ?

luakit version: c5328c22

c0dev0id:
Hi,

it looks like you're using a pretty old release. Maybe consider updating to version 2.1.

This works for me and I think it also worked in the August 2017 release...:

```
-- Use "qwertzuiopasfghjklyxcvbnm" for generating labels
local select = require "select"
select.label_maker = function (s)
    return s.trim(s.sort(s.reverse(s.charset("qwertzuiopasfghjklyxcvbnm"))))
end
follow.pattern_maker = follow.pattern_styles.match_label
```

delister:
Thx, update to 2.1 helped me.

PR #1038, opened 2023-02-28
docs: demonstrate that `label_maker` can accept a table

fosskers:
As-is, the existing examples fail in certain environments. By passing in the table of `label_styles`, the "styling functions" can be accessed directly.

See also:

- https://github.com/luakit/luakit/issues/782
- https://github.com/bakpakin/Fennel/issues/443

### DIFF
--- a/lib/follow.lua
+++ b/lib/follow.lua
@@ -19,9 +19,9 @@
 --
 --     local select = require "select"
 --
---     select.label_maker = function ()
---         local chars = charset("asdfqwerzxcv")
---         return trim(sort(reverse(chars)))
+--     select.label_maker = function (s)
+--         local chars = s.charset("asdfqwerzxcv")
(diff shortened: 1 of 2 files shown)


fosskers:
Hi there, any thoughts?

taobert:
Hi there.
Thankyou for the PR, and sorry for the lack of response.
Luakit has a very small set of developers working in their spare time, 
and even tasks like vetting PRs can take quite a while to get to,
depending on other real-world commitments.

I can confirm that the change you propose to the docs is consistent with what i use,
and from the looks of #782 which you link to, also @c0dev0id.

I'm not personally able to accept the PR.
If i were, i'd want to try to find the time to figure out exactly when and how the source and docs diverged,
just to confirm that there are no other related gotchas that should also be mentioned/corrected.

fosskers:
Thanks!

PR #1040, opened 2023-04-27
Fix "package.path is not a string" error on latest LuaJIT

PluMGMK:
When luakit is compiled against the current v2.1 branch of LuaJIT, it fails to start because "package.path is not a string", so it can't add the needed searchpaths for Lua modules. (This can be worked around by setting the `LUA_PATH` environment variable, but that's far from being an ideal solution.)

Apparently, the function `luaH_add_paths` assumed that "package" would always be in position 1 on the stack after calling `lua_getglobal`, which was valid on LuaJIT 2.0.5, but (for some reason I can't explain) doesn't hold on the newer version. It seems the fix is simply to use relative offsets from the stack top instead, i.e. pass negative numbers to `lua_getfield`, `lua_type` and `lua_setfield`.

I think this qualifies as a "tweak needed to get luakit to run"; that said, I should note that I know very little about Lua (just the bare minimum to configure luakit itself), and I knew nothing about the Lua C API until I started digging into this yesterday…

### DIFF
--- a/common/luautil.c
+++ b/common/luautil.c
@@ -141,12 +141,12 @@ void
 luaH_add_paths(lua_State *L, const gchar *config_dir)
 {
     lua_getglobal(L, "package");
-    if(LUA_TTABLE != lua_type(L, 1)) {
+    if(LUA_TTABLE != lua_type(L, -1)) {
         warn("package is not a table");
         return;
     }
-    lua_getfield(L, 1, "path");
-    if(LUA_TSTRING != lua_type(L, 2)) {
+    lua_getfield(L, -1, "path");
+    if(LUA_TSTRING != lua_type(L, -1)) {
         warn("package.path is not a string");
         lua_pop(L, 1);
         return;
@@ -193,7 +193,7 @@ luaH_add_paths(lua_State *L, const gchar *config_dir)
     g_ptr_array_free(paths, TRUE);
 
     /* package.path = "concatenated string" */
-    lua_setfield(L, 1, "path");
+    lua_setfield(L, -2, "path");
 
     /* remove package module from stack */
     lua_pop(L, 1);


taobert:
I may be wrong, but i suspect there may be more going on here.
Debian (Sid) ships `LuaJIT 2.1.0-beta3`, and luakit functions with it.
This package is apparently built from the [openresty branch](https://github.com/openresty/luajit2) rather than the main luajit repo, so that may be pertinant.

PluMGMK:
OK, I can confirm that luakit also works for me without this patch when using the openresty branch. I wonder what's going on there…

That said, I think it generally makes more sense to use relative offsets on the stack, doesn't it?

taobert:
I can't comment on whether using relative offsets on the stack is more sensible,
though i feel that the code base should be kept as consistent as possible.
I haven't investigated, but i would guess that if non-relative offsets are used here, they're probably used elsewhere. 

In my hands, i see this failure-to-start if i use the luajit bleading edge (`git checkout v2.1`)
but not if i use (what i believe is) the previous commit (`git checkout v2.1~1`).
It may be that this is just a short lived error that will be remedied soon.
The last commit doesn't obviously (to me) cause this behavior (i gather it's changing error reporting in garbage collection).
I don't see a relevant-looking bug reported on their github page, so you might want to clarify the situation there.
If they confirm that the new behaviour is indeed prefered, then i guess we need to work with it.

PluMGMK:
> I haven't investigated, but i would guess that if non-relative offsets are used here, they're probably used elsewhere.

Agreed, although it is interesting that this seems to be the only breakage in the entire codebase…

> In my hands, i see this failure-to-start if i use the luajit bleading edge (`git checkout v2.1`) but not if i use (what i believe is) the previous commit (`git checkout v2.1~1`). It may be that this is just a short lived error that will be remedied soon. The last commit doesn't obviously (to me) cause this behavior (i gather it's changing error reporting in garbage collection). I don't see a relevant-looking bug reported on their github page, so you might want to clarify the situation there. If they confirm that the new behaviour is indeed prefered, then i guess we need to work with it.

I suppose I'll go and do that. For what it's worth, it seems the "extra" value at position 1 on the stack is an empty table, not sure how it's getting there.

PluMGMK:
OK, the problem is fixed in LuaJIT now, but the author still thinks this change is a good idea. Maybe I'll have a look and see if absolute stack offsets are used anywhere else…

PR #1047, opened 2023-07-11
add otherwise style rule

ghost:
Add an "otherwise" @-moz-document rule to be used when no other stylesheet matches. "otherwise" css files will be disabled by any matching stylesheet. The argument is ignored.

example:
@-moz-document otherwise('') { }

### DIFF
--- a/lib/styles.lua
+++ b/lib/styles.lua
@@ -113,6 +113,8 @@ end
 -- Routines to re-apply all stylesheets to a given webview
 
 local function update_stylesheet_application(view, domains, stylesheet, enabled)
+    local anymatch
+
     for _, part in ipairs(stylesheet.parts) do
         local match = false
         for _, w in ipairs(part.when) do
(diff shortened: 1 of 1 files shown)


ghost:
This works, but it's not pretty. I welcome any suggestions on how to do it better.

lib/styles.lua
```diff
@@ -125,6 +127,21 @@ local function update_stylesheet_application(view, domains, stylesheet, enabled)
...
         end
+        anymatch = anymatch or match
+        match = match and stylesheet.enabled and enabled
+        view.stylesheets[part.ss] = match
+    end
+
+    return anymatch
+end
+
+local function update_stylesheet_application_otherwise(view, domains, stylesheet, enabled)
```
c0dev0id, line 138: You're passing "domains", but are not using it. This makes the test fail:
```
FAIL style/test_luacheck / test_luacheck
  tests/style/test_luacheck.lua:103: Luacheck messages:
    lib/styles.lua:138: (62:68):           unused argument 'domains'
```

lib/styles.lua
```diff
@@ -133,11 +150,18 @@ end
...
 
 local function update_stylesheet_applications(v)
+    local anymatch
     local enabled = v:emit_signal("enable-styles")
     enabled = enabled ~= false and true
     local domains = domains_from_uri(v.uri)
     for _, s in ipairs(stylesheets or {}) do
-        update_stylesheet_application(v, domains, s, enabled ~= false )
+        local a = update_stylesheet_application(v, domains, s, enabled ~= false )
+        anymatch = anymatch or a
```
c0dev0id, line 159: The temporary variable is not needed here.
```
anymatch = anymatch or update_stylesheet_application(v, domains, s, enabled ~= false )
```

c0dev0id:
Sorry, I added my review long ago, but forgot to push "submit review". So it was visible for me, but in pending state.

Issue #1052, opened 2023-07-31
New webkit renders PDF itself


msdemlei:
**Current Behavior:**

When luakit "opens" a PDF, it (well: the embedded webkit) renders it itself rather than handing it over to viewpdf or just downloading it.  That behaviour was introduced somewhere between webkit 2.34 and 2.40.

**Desired Behavior:**

Have a knob (in settings?) to turn off the webkit-internal PDF renderer (if you ask me: I'd make it default off).

**How can we reproduce it (step by step):**

Link against webkit2gtk 2.40.3 (as in Debian bookworm), follow a link to an application/pdf.

**Environment:**

Linux Distribution & Version: Debian bookworm
Output of `luakit --version`:
luakit 2.3.3
  built with webkit 2.38.2 (installed version: 2.40.3)

[Hm... now that I read this... could this be trouble?]

taobert:
> [Hm... now that I read this... could this be trouble?]

No, i see the behaviour too with:
```
# luakit --version
luakit 2.3.3-40-g73ce8add
  built with: webkit 2.40.4 (installed version: 2.40.4)
                 GTK 3.24.38
                GLIB 2.76.4
                SOUP 3.2.2
```

henkmet:
The good thing is that webkit now renders the pdf correctly whereas earlier it would be a sure crash (for me). But likewise, I find that I'm very often downloading the file to use my system viewer. A seting would be welcome.

msdemlei:
I've dug a bit, and over at WebKit, there's even a PR tag for the whole calamity: https://github.com/WebKit/WebKit/pulls?page=2&q=is%3Apr+pdf+is%3Aclosed+label%3APDF, but there are quite a few more PRs that mention "PDF" on top.  The one introducing pdfjs in the first place, however, I've not found yet (but then I was was too stingy to clone the whole tree, which looks like several gigabyte-ish).

There's an easy-to-find build-time setting to throw out pdfjs, but I certainly don't want to rebuild webkit (regularly); the code that decides to do some wild iframing for application/pdf, on the other hand, I have not found.

I've also experimented with blocking the webkit-pdfjs-viewer: URI scheme webkit uses internally from within luakit, but that (understandably) results in a white page when clicking a PDF.

So... I guess I'd rather ask around rather than keep on poking blindly.  Does anyone know a good place to ask?

PR #1069, opened 2024-01-07
Add pdfjs switch

msdemlei:
This is an attempt to address bug #1052; I'm (well: my machine is) tired of rebuilding luakit to patch out pdf.js.

What this does now is follow Michael's recipe on https://bugs.webkit.org/show_bug.cgi?id=260019#c4 (comment 4) to unconditionally turn off pdfjs using some ugly custom code in widgets/webview.c; that works on a PoC level.

However, assuming that not everyone is as abhorred by pdf.js as I am, this would now need to be made configurable.  I *think* we will still need C code for this, as webkit_settings_get_all_features does not seem to be exposed to lua (and I don't think it should). But I have not found a way to trick that C code into settings.lua; actually, I have not even worked out how settings.lua talks to webkit in "normal" cases.

Does anyone here still understand settings.lua?  Well enough to figure out whether we can have a setting that simply calls a C function on toggling, even?

### DIFF
--- a/lib/webview.lua
+++ b/lib/webview.lua
@@ -575,7 +575,7 @@ local webview_settings = {
     },
     ["application.enable_pdfjs"] = {
         type = "boolean",
-        default = false,
+        default = true,
         desc = [=[
             Whether to load PDFs in a built-in javascript renderer.
             The default is to pass them on to the OS.  This setting
--- a/lib/webview.lua
+++ b/lib/webview.lua
@@ -145,6 +145,12 @@ local init_funcs = {
             end
         end)
     end,
+
+    enable_pdfjs = function (view)
+        local pdfjs_enabled =
+            settings.get_setting("application.enable_pdfjs")
+        view:set_pdfjs(pdfjs_enabled or false)
+    end,
 }
 
 --- These methods are present when you index a window instance and no window
@@ -567,6 +573,14 @@ local webview_settings = {
             Disabling this setting is only useful to conserve memory.
         ]=],
     },
+    ["application.enable_pdfjs"] = {
+        type = "boolean",
+        default = false,
+        desc = [=[
+            Whether to load PDFs in a built-in javascript renderer.
+            The default is to pass them on to the OS.  This setting
+            only becomes active in new tabs.]=]
+    },
     ["webview.enable_plugins"] = {
         type = "boolean",
         default = true,
@@ -717,8 +731,9 @@ settings.register_settings({
 
 _M.add_signal("init", function (view)
     local set = function (wv, k, v, match)
-        if v ~= nil then
-            k = k:sub(9) -- Strip off "webview." prefix
+        -- hand through webview.-prefixed settings
+        if v ~= nil and k:find('webview.', 1, true) then
+            k = k:sub(9) -- Strip off prefix
             if k == "zoom_level" then v = v/100.0 end
             if k == "user_agent" and v == "" then v = nil end
             match = match and (" (matched '"..match.."')") or ""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Manage dom events with luakit signals.
+- An enable_pdfjs setting to go back to letting viewpdf handle PDFs.
 
 ### Changed
 
--- a/common/tokenize.list
+++ b/common/tokenize.list
@@ -124,6 +124,7 @@ serif_font_family
 send_key
 set_dark_mode
 set_default_size
+set_pdfjs
 set_title
 show
 show_border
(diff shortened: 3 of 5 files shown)


c0dev0id:
You don't need to touch `settings.lua`. You can register a toggle like I have done it when adding dark-mode support here: https://github.com/luakit/luakit/commit/ea082a5b496b1ca2db80c676e0976d9d082c9295

Except your are registering your toggle in `webview.lua` and read it from `webview.c`.
There are already settings being registered in `webview.lua`. You can add a toggle there.

msdemlei:
On Tue, Jan 23, 2024 at 02:46:19PM -0800, Stefan Hagen wrote:
> You don't need to touch settings.lua. You can register a toggle like I have done it when adding dark-mode support here: https://github.com/luakit/luakit/commit/ea082a5b496b1ca2db80c676e0976d9d082c9295

Thanks for the hints -- that's really helpful, although of course I'm
basically blindly prodding along, wondering if there is a place I can
read up on all that (well: the lua-C interface I know where to read
up on, and I'll do that now).

But I've reached a place where I'll need someone with actual
knowledge about the interface to webkitgtk: It seems I can't just
invent configuration keys in webview.lua, and I wonder if I'm just
being dense or whether I really need to do this some other way.

You see, webview.lua has a signal handler for init which executes
set_all.  That, in turn, iterates over webview_settings, drops the
constant prefix "webview" and then runs wv[config-key] = somevalue.

With my new setting webview.enable_pdfjs, that fails with:

  Lua error: cannot set unknown webview property 'enable_pdfjs'

So...

> There are already settings being registered in webview.lua. You can
> add a toggle there.

Is it actually right to just add something in webview_settings or
do these have to reflect properties of the C-side object?  And if I
can bother you to have a quick look at the last commit: Does the rest
look as if it could work?

msdemlei:
So, I'm now creating an application.enable_pdfjs setting in webview.lua, which is the first application setting in there.  To keep that out of libwebkitgtk's view, I'm now doing prefix checking in the init signal handler.

I *hope* this is ready for review now.  One thing someone knowledgeable should definitely look at: do I need to free any of what comes back from the webkit calls in luaH_webview_set_pdfjs except for the result of webkit_settings_get_all_features (where the docs say I own it)?  And is it ok to use g_autoptr in luakit?  Glib docs say it's only available in gcc and clang – is this something we want to avoid?

lib/webview.lua
```diff
@@ -567,6 +573,14 @@ local webview_settings = {
             Disabling this setting is only useful to conserve memory.
         ]=],
     },
+    ["application.enable_pdfjs"] = {
+        type = "boolean",
+        default = false,
+        desc = [=[
+            Whether to load PDFs in a built-in javascript renderer.
+            The default is to pass them on to the OS.  This setting
+            only becomes active in new tabs.]=]
```
c0dev0id: I'm a bit torn weather we should enable or disable it per default.

Without your PR, webkit renders PDFs itself. So your PR changes this behavior.

But before webkit started display PDFs, a download dialog was shown and we had the [viewpdf](https://luakit.github.io/docs/modules/viewpdf.html) module to handle it.

I tend to enabling pdfjs per default and adding a sentence to the viewpdf documentation that it requires this setting to be toggled off to work.

lib/webview.lua
```diff
@@ -717,8 +731,9 @@ settings.register_settings({
 
 _M.add_signal("init", function (view)
     local set = function (wv, k, v, match)
-        if v ~= nil then
-            k = k:sub(9) -- Strip off "webview." prefix
+        -- hand through webview.-prefixed settings
+        if v ~= nil and k:find('webview.', 1, true) then
+            k = k:sub(9) -- Strip off prefix
```
c0dev0id, line 736: So... it looks like this is the first setting that's retrieved from the webview, but it's actually an application setting. My first thought was that this should go into `window.lua`/`window.c`. But then we need to include webview there, which is wrong. Therefore I think this is ok.

c0dev0id:
> do I need to free any of what comes back from the webkit calls in luaH_webview_set_pdfjs except for the result of webkit_settings_get_all_features

I'm not sure. The memory leak unit test does not complain, so I assume it's okay as it is.

> g_autoptr

We can try it. And we change it in case it creates trouble.

msdemlei:
On Tue, Feb 06, 2024 at 02:11:40AM -0800, Stefan Hagen wrote:
> I tend to enabling pdfjs per default and adding a sentence to the
> viewpdf documentation that it requires this setting to be toggled
> off to work.

Ah...  well.  I *would* have bet money that the clientele that uses
luakit would howl about their PDFs not opening in their beloved
zathura (or whatever), but since nobody but me gnashed their teeth
I'm willing to accept that I may be slightly odd.  So, in aca1ee5a
I'm turning pdfjs on by default.  I'm also making a brief note in the
changelog in that commit.

For now I can't think of anything else to do here, so from where I
stand this looks ready to go.

Thanks for the review!

PR #1073, opened 2024-01-30
Correct regex compile options type if none are passed

jpalus:
compile options are of enum type so don't try to pass pointer instead. fixes compile failures (GCC13):

log.c:225:46: error: incompatible type for argument 2 of 'g_regex_new'
./common/clib/soup.h:164:59: error: incompatible type for argument 2 of 'g_regex_new

Fixes: e400cda67c81 ("Remove G_REGEX_OPTIMIZE")


c0dev0id:
Fixing the fix. Thank you!

PR #1074, opened 2024-02-01
Don't compress man pages

sideeffect42:
Compressing man pages should be done by the package manager based on distribution policy or user configuration.

On (at least) Gentoo the installation of pre-compressed man pages results in a QA error.

https://cmpct.info/~sam/blog/posts/automatic-manpage-compression/


c0dev0id:
I'm sharing this sentiment and also patched it out on OpenBSD. Appearently debian wants it compressed? https://github.com/luakit/luakit/commit/fad4541a13492566a8b49b8291e3668051f74b66

I agree that this should be left to package maintainers. Thanks for the PR.

PR #1076, opened 2024-02-08
Install missing dependencies for tests.yml

aidanholm:


c0dev0id:
Oh if the CI setup can run again, that would be amazing.

aidanholm:
Should be ready :)

This only covers the luajit + ubuntu combo, would be good to get some more up, but at least the tests will run with every commit / PR

clib/unique.c
```diff
@@ -78,7 +78,7 @@ luaH_unique_new(lua_State *L)
 
     GError *error = NULL;
     if (!globalconf.application) {
-#if GTK_CHECK_VERSION(2,74,0)
+#if GLIB_CHECK_VERSION(2,74,0)
```
c0dev0id, line 81: Unrelated to the PR, but good catch.

.github/workflows/tests.yml
```diff
@@ -13,5 +13,16 @@ jobs:
...
     - uses: actions/checkout@v3
+    - name: Install depenencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install luajit libluajit-5.1-dev libjavascriptcoregtk-4.0-dev libwebkit2gtk-4.1-dev libgtk-3-dev libglib2.0-dev luarocks
+    - name: Install lua dependencies
+      run: |
+        sudo luarocks --lua-version 5.1 install luafilesystem
+        sudo luarocks --lua-version 5.1 install luassert
+        sudo luarocks --lua-version 5.1 install luacheck
```
c0dev0id, line 24: Why from luarocks and not via apt? `lua-check lua-luassert lua-filesystem` are available for lua 5.1 and my debian machine here. So I assume they are available here too. Or not?

c0dev0id (approved):

.github/workflows/tests.yml
aidanholm: Ah, the reason is that I simply forgot about that possibility :) sent https://github.com/luakit/luakit/pull/1077

PR #1077, opened 2024-02-08
Install lua dependencies via apt instead of luarocks.

aidanholm:
This avoids installing luarocks, which has a few extra deps of its own.

These apt packages contain four copies, for lua 5.1 through 5.4, but that should be much less expensive.


c0dev0id:
Yep that's better. It also resembles the setup that debian/ubuntu users have.
Thanks!

PR #1083, opened 2024-03-31
Fixed project URL in quick start guide

rdbo:
I was following the quick start guide, and I noticed that it points to the URL `luakit.org`, which doesn't seem to be related to this project.
I modified the URL to `luakit.github.io`, which is the main page for the project.


c0dev0id:
Thanks!

PR #1088, opened 2024-09-15
noscript: update policy on the navigation-request signal

### LINKED ISSUE #649: noscript domain matching issues
There are a few scenarios where noscript seems to act strangely with domain overrides (unexpected behaviour bolded):

1. `:open` seems to match against the previous open url/domain
    1. `noscript.enable_scripts = false`
    2. starting from `about:blank`
    3. `:open domain.com` where `domain.com` is set to override and allow scripts
    4. **`domain.com` opens without executing scripts**
    5. `:reload`
    6. scripts now work
2. `,ts` toggle scripts on seems to require two refreshes to take effect
    1. `noscript.enable_scripts = false`
    2. navigate to site that uses javascript but has no overrides set
    3. thus no scripts execute
    4. `,ts` to toggle scripts on
    5. `:reload`
    6. **page loads again without executing scripts**
    7. `:reload`
    8. scripts now work

balejk:
Replace the policy-decided signal which is never emitted with
navigation-request. This should resolve some JavaScript policy issues
such as the page needing to be refreshed for the correct policy to take
effect.

---
Fixes: #649

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 - Fixed undoclose bug (and likely some other segfaults)
 - Fixed luaH_init() implicit prototype warning
+- Fixed refresh being needed for the correct NoScript policy to take effect.
 
 ### Contributors to this release:
 
--- a/lib/noscript.lua
+++ b/lib/noscript.lua
@@ -197,10 +197,8 @@ end
 
 webview.add_signal("init", function (view)
     -- Update on new resource load
-    view:add_signal("policy-decided", function (v, _, _, decision)
-        if decision == "use" then
-            update_webview_blocking(v)
-        end
+    view:add_signal("navigation-request", function (v, _, _)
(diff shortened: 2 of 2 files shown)


balejk:
It seems that I'm still able to hit the "refresh needed" problem if I have `webview.enable_javascript = false` in my `userconf.lua`, but since `noscript` has its own default I supposed mixing these is not supported anyway (also see the first reply in the referenced issue)?

Also, I have started to have some doubts on whether the `navigation-request` signal is really the right one for this, so I will certainly appreciate tips if there is a better one. Nevertheless, the fact that `policy-decided` is never emitted is I believe most certainly a bug and this does fix the issues I have been having with `noscript` (with only the above caveat noticed so far), so I think this should certainly still be a step forward :-)

c0dev0id:
I'm personally not using noscript, but this seems to work as advertised.

PR #1089, opened 2024-09-15
follow: add <summary> as a clickable element

balejk:
Clicking this element will toggle the visibility of the content of the
containing <details> element.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Manage dom events with luakit signals.
 - An enable_pdfjs setting to go back to letting viewpdf handle PDFs.
+- `<summary>` elements are now hinted as clickable.
 
 ### Changed
 
--- a/lib/follow.lua
+++ b/lib/follow.lua
@@ -342,7 +342,7 @@ add_binds("follow", {
 -- @type {[string]=string}
 -- @readwrite
 _M.selectors = {
-    clickable = 'a, area, textarea, select, input:not([type=hidden]), button, label',
+    clickable = 'a, area, textarea, select, input:not([type=hidden]), button, label, summary',
     -- Elements that can be clicked.
     focus = 'a, area, textarea, select, input:not([type=hidden]), button, body, applet, object',
     -- Elements that can be given input focus.


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Manage dom events with luakit signals.
 - An enable_pdfjs setting to go back to letting viewpdf handle PDFs.
+- The `<summary>` element to clickable selectors in the `follow` plugin.
```
msdemlei: This is not quite a proper sentence.  Suggestion: "summary elements can to be operated through f hints, too".

msdemlei (changes requested):
Thanks for this contribution!

I think having summary keyboard-operatable is right.  However, I had to look it up, too: The idea is that you toggle the visibility of an associated details in this way.

Perhaps you can state about as much in the commit message? Other than that I'd say let's go.

balejk:
> I think having summary keyboard-operatable is right.  However, I had to look it up, too: The idea is that you toggle the visibility of an associated details in this way.
>
> Perhaps you can state about as much in the commit message? Other than that I'd say let's go.

I'm confused, isn't that exactly what I wrote there?

> > @@ -6,6 +6,7 @@
>  
>  - Manage dom events with luakit signals.
>  - An enable_pdfjs setting to go back to letting viewpdf handle PDFs.
> +- The `<summary>` element to clickable selectors in the `follow` plugin.
>
> This is not quite a proper sentence.  Suggestion: "summary elements can to be operated through f hints, too".

Well, I followed your precendence a line above :-) But if you still
think it should be changed, I can certainly reword it. How about
"<summary> elements are now hinted as clickable."?

Thank you for your feedback!

msdemlei:
On Tue, Sep 17, 2024 at 01:42:29PM -0700, balejk wrote:
> > I think having summary keyboard-operatable is right.  However, I had to look it up, too: The idea is that you toggle the visibility of an associated details in this way.
> >
> > Perhaps you can state about as much in the commit message? Other than that I'd say let's go.
>
> I'm confused, isn't that exactly what I wrote there?

You did, sorry, it was me who was confused.

> > This is not quite a proper sentence.  Suggestion: "summary elements can to be operated through f hints, too".
>
> Well, I followed your precendence a line above :-) But if you still
> think it should be changed, I can certainly reword it. How about
> "<summary> elements are now hinted as clickable."?

That's great.  Let's go for it (ideally with a commit --amend and a
push -f).  And yes, I notice my changelog entry was not a good
sentence either.  Should be fixed before the next release, sure.

Thanks!

balejk:
I have pushed the change.

c0dev0id:
Looks good. I'll merge it. Thanks!

PR #1091, opened 2024-09-29
[RFC] web_context: remove deprecated WebsiteDataManager properties

balejk:
Also rename the corresponding function to better reflect what it does.

---

I believe it should be fairly easy to automatically create symbolic links to have the new locations point to the old ones iff those exist on luakit startup, however I have not implemented it yet as I'm not sure whether it is desirable to make such an operation part of the program instead of requiring a one-time manual intervention explained in the changelog (it would certainly be much smoother and comfortable for the user though) -- please let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@
 
 - Don't compress man page. This should be done by maintainers.
 - Removed now unsupported/ignored load-icons-ignoring-image-load-setting.
+- The default WebKit data and cache subdirectories are now used. If you use any
+  website which uses IndexedDB or local storage and wish to retain the saved
+  information, you will need to manually move the corresponding directories to
+  their new locations. Specifically, the mapping is as follows:
+
+  - `$XDG_DATA_HOME/luakit/local_storage -> $XDG_DATA_HOME/luakit/localstorage`
+  - `$XDG_DATA_HOME/luakit/indexeddb -> $XDG_DATA_HOME/luakit/databases/indexeddb`
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


c0dev0id:
Thank you. Fixes for deprecations are very welcome. I'm also working on a few of those in order to upgrade the webkitgtk version. I'll test your PRs in the next days and merge them if nothing comes up.

balejk:
> I'm also working on a few of those in order to upgrade the webkitgtk version.

You mean to upgrade to webkit-6.0 and gtk4? That was my motivation for
these two PRs too. From what I have been able to tell, after this it
remains to stop using the deprecated JavaScriptCore API in favour of the
new one and rewrite the DOM functionality in JavaScript. Are you working
on these then?

c0dev0id:
Frankly, I don't think I can pull off a full migration. I started working on fixing migrations in the current version.  I'm currently looking at replacing the deprecated gtk menu system with the gio menu system. Which lead me down the rabbit hole of learning the lua C API (which I probably should properly learn anyway).

balejk:
I have been skimming through the Lua C API (and WebKit, GTK, JSC and
GLib/GObject) docs myself when looking into the JSC rewrite. It seems to
me that it shouldn't be too hard for someone who knows what they are
doing which I'm sadly not at the moment.

By the way, you haven't replied to my question regarding the change of
directories, do you thus think the changelog entry is sufficient? I just
want to make sure because I believe there can be cases where the loss of
the old locations could be very inconvenient, for instance I suspect the
Matrix web-client Element uses one of those locations to store the
message encryption and identity keys.

c0dev0id:
Yes, I think it's sufficient. I'll do version 2.4 next and highlight breaking changes on the release page. I should probably have done this last release already with the move from libsoup2 to libsoup3. That's why I provided a libsoup2 version as well. That's not going to happen with the next release.

PR #1094, opened 2024-10-20
luajs: replace luaJS_eval_js with a variant using the new JSC API

balejk:
- **webview: remove unused JavaScriptCore include**
- **luajs: replace luaJS_eval_js with a variant using the new JSC API**


c0dev0id:
Hi balejk,

that's cool, thank you. I see a few tests failing. It's interesting to see test failing that have been a bit unstable in the past. Especially the undoclose one. It could be that your change is correct, but made the behavior more deterministically failing.

I'll test it in the next days.

balejk:
> I'll test it in the next days.

Thank you and sorry about not running the test suite, my distribution
doesn't package luassert which seems to be required for that. I will try
to get it installed another way.

I have converted this to draft for now because I want to make sure I'm
freeing the JSC objects correctly. I will mark it ready as review once I
do that so please don't merge before that.

I am also slightly unsure that I haven't broken luaJS_pushobject for
numeric indices. I didn't originally find a way to enumerate those on
the JSC JS object, however now I'm wondering whether treating the object
as an array could work, although I don't see a method for enumerating
those either.

Comments would be very welcome.

balejk:
I believe I have fixed the memory issues. I have also improved some
other parts a little bit. I still plan to look into the array
enumeration.

Unfortunately though, I haven't been successful in getting the test
suite working with luassert installed via luarocks. I have exported the
variables produced by `luarocks path` and the test suite does run which
it did not before, however all the relevant tests themselves still seem
to fail on require("luassert").

balejk:
Sorry for the delay and confusion, I have been looking at the array
problem perhaps in a too complicated way. I have now simply modified
this to do what the old implementation (of luaJS_pushobject) did (or
actually still does).

While doing that, I noticed that unless I'm mistaken, the old
implementation doesn't handle object keys which are an empty string
correctly because in such case cstr points to NUL and strtol will make
eptr point to the same place, hence making it satisfy the if condition
without cstr actually being a valid number. If I'm understanding
correctly, the pushed integer would then be 0 which doesn't seem wanted.
Please correct me if I'm wrong.

I believe this should be ready now.

balejk:
I have updated this with the alternative approach to `page:wrap_js` update mentioned in #1095 not to open a third PR with a similar content, I hope that is OK, otherwise please let me know and I can revert this back to the previous state.

Personally I prefer the approach implemented here, it feels to me that this functionality makes sense for `eval_js` and that there is no need to separate this out into a dedicated function, however the choice is up to you of course, feel free to close/merge either of these two PRs (but ideally not both obviously :-). And of course, if there is anything that I should improve in either of them, please let me know.

balejk:
So similarly to what I wrote at the other PR, I would still like to make some improvements here before it's merged, in particular concatenate the wrapped script directly on the Lua stack instead of using strcat which should be nicer, so I have converted this back to draft and will implement this once you let me know which of the PRs you prefer.

balejk:
@c0dev0id I have updated this per the above comment and our IRC discussion (please let me know if I forgot something) and I think it is ready for a review.

c0dev0id:
Hmm. It's interesting that the undoclose test is failing reliably now on github. Running the test suite locally works fine for me. The code looks fine to me. I'll do a "rebase and merge" now. I think this should have the expected result of bringing the commits 1:1 into the develop branch.

PR #1095, opened 2024-10-29
[RFC] rewrite page:wrap_js() using the new JSC API and expose it

balejk:
- **webview: remove unused JavaScriptCore include**
- **luajs: replace luaJS_eval_js with a variant using the new JSC API**
- **page: use the new JSC API for wrap_js**
- **select_wm: do not unnecessarily re-evaluate client_rects**
- **doc: document page:wrap_js()**

--

Builds atop #1094.

Please do not merge this just yet, it still occurred to me that it might be possible and nice to provide this functionality via the existing `page:eval_js`. I am just putting this up for comments and review in the meantime, before I have time to look into that idea.

In particular, I'm mainly looking for reviews of the GObject JSC rewrite patch and for opinions on whether the two following patches are appropriate.

Still with regards to the rewrite, I believe it would be possible to keep the usage of `page:wrap_js` the same as before by wrapping the provided JS function body with the necessary additions in C, I'm just not sure that it's actually worth it. However, if I manage to incorporate this into `page:eval_js`, this could be left around only as a simple wrapper for backwards compatibility, doing just that and passing it to `page:eval_js`. I will be very happy for opinions on this as well as for any other feedback.


balejk:
I would still like to make a few cosmetic changes here, mainly use anonymous JS functions in the wrapped code snippets, so I'm leaving this as a draft and please let me know if you choose this PR over the other one in which case I will implement and push these before it is merged.

balejk:
Closing in favour of #1094.

Issue #1067, opened 2023-12-21
coredumps


henkmet:
**Current Behavior:**
After updating webkit to 2-42-4 on archlinux (and rebuilding luakit) I'm experiencing instabilities. Two times a SIGSEGV and once a SIGABRT

**Desired Behavior:**
don't crash

**How can we reproduce it (step by step):**
As before, it seems archlinux.org triggers something. I'd need to do some more testing for step-by-step reproduction.

**Environment:**

Linux Distribution & Version:
archlinux, linux 6.6.7.arch1-1
Output of `luakit --version`:
luakit 2.3.3-15-gb143b383
  built with: webkit 2.42.4 (installed version: 2.42.4)
                 GTK 3.24.38 
                GLIB 2.78.3 
                SOUP 3.4.4 

**Note about webkit issues:**

If you're reporting a rendering issue, please test it with the gnome
browser ephiphany as well. If the issue occurs there too, we're very
likely not able to help. These issues should be reported to webkit:
https://bugs.webkit.org

I don't know how useful backtraces are but there seems to be an issue with malloc? (This is from the SIGABRT coredump)
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007fa4dfefe8a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007fa4dfeae668 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fa4dfe964b8 in __GI_abort () at abort.c:79
#4  0x00007fa4dfe97390 in __libc_message (fmt=fmt@entry=0x7fa4e000e55d "%s\n") at ../sysdeps/posix/libc_fatal.c:150
#5  0x00007fa4dff087b7 in malloc_printerr (str=str@entry=0x7fa4e00117d8 "malloc(): mismatching next->prev_size (unsorted)") at malloc.c:5765
#6  0x00007fa4dff0bd9c in _int_malloc (av=av@entry=0x7fa438000030, bytes=bytes@entry=88) at malloc.c:4076
#7  0x00007fa4dff0ccad in __GI___libc_malloc (bytes=bytes@entry=88) at malloc.c:3329
#8  0x00007fa4e00b4763 in g_malloc (n_bytes=88) at ../glib/glib/gmem.c:130
[...]

PR #1109, opened 2025-04-10
GTK Native File Chooser support

### LINKED ISSUE #1107: Use filechooser set by xdg desktop portal
Is there a way to make Luakit use the filechooser specified by xdg desktop portal for downloads? 

When I download a file in luakit the gtk file chooser is always opened. 
However, when I upload a file the correct filechooser is used. 

Thanks.

**Environment:**

Linux Distribution & Version: gentoo 
Output of `luakit --version`:
```
luakit 636637cd
  built with: webkit 2.46.5 (installed version: 2.46.5)
                 GTK 3.24.49 
                GLIB 2.82.5 
                SOUP 3.4.4 
```

fictitiousexistence:
Add support for using the File Chooser set by XDG Desktop Portal via    
"GTK_USE_PORTAL" environment variable.                                  
                                                                        
Users can change which File Chooser is used instead of using the default 
gtk chooser. For example, users can use kde's file chooser or even custom
choosers like: boydaihungst/xdg-desktop-portal-termfilechooser          
                                                                        
If GTK_USE_PORTAL is set to 0 or no alternate chooser is set the default 
gtk chooser will be used.                                               
                                                                        
Tested on sway,dwl,gnome,kde                                            
                                                                        
resolves #1107


fictitiousexistence:
Before applying this I noticed that after I downloaded anything I would get spammed with thousands of `[core/common/ipc]: Trying to send an ipc message, but the endpoint went away.` messages. This has not seemed to happen today.

balejk:
The error messages you are seeing are due to #1067, so they should not be related to your PR.

I haven't reviewed the actual code, just a nit: you seem to be making an unnecessary newline change in the middle hunk.

PR #1110, opened 2025-04-11
Fix Download Remove from list

### LINKED ISSUE #1108: Download issues
Hi, I ran into some issues with downloading files in luakit.

1.) 

If I go to download a file but decide to click cancel on the save as dialog I get the following error:

```
Lua error: attempt to index field '?' (a nil value)
Traceback:
(1) /usr/share/luakit/lib/downloads.lua:204 in function [anonymous]
```

In /usr/share/luakit/lib/downloads.lua I changed:
```
    d:add_signal("finished", function(dd)
        query_insert:exec{os.time(), dls[dd].created, dd.uri, dd.destination, dd.total_size}
```
to 
```
    d:add_signal("finished", function(dd)
        if not dd.destination then return end                                         
        query_insert:exec{os.time(), dls[dd].created, dd.uri, dd.destination, dd.total_size}
```
and that seems to fix the issue. I'm not a programmer so I don't know if this is right.  

2.)

If I download a file using the :download command, it throws this error but the file is still downloaded:
```
Lua error: sqlite3: exec error (UNIQUE constraint failed: downloads.finished_time)
Traceback:
(1) [C]                                     in function exec
(2) /usr/share/luakit/lib/downloads.lua:204 in function [anonymous]
```

I changed /usr/share/luakit/lib/downloads.lua:204 from:
`query_insert:exec{os.time(), dls[dd].created, dd.uri, dd.destination, dd.total_size}`
to 
`query_insert:exec{os.time()..math.random(1,99), dls[dd].created, dd.uri, dd.destination, dd.total_size}` 
to add some extra randomness and that stopped the error and then I realised the error is occuring because the :download command
trys to update the downloads.db twice.

If I download a file via ex: right click /save as, it does not duplicate it in downloads.db  

3.)

The "Remove from list" on luakit://downloads/ does not work. 
When clicked it will remove it from the list on the page but not from downloads.db. 
After restarting luakit the downloads will show back up in luakit://downloads/. 

**Environment:**

Linux Distribution & Version: gentoo 
Output of `luakit --version`:
luakit 636637cd
  built with: webkit 2.46.5 (installed version: 2.46.5)
                 GTK 3.24.49 
                GLIB 2.82.5 
                SOUP 3.4.4

fictitiousexistence:
The "Remove from list" on luakit://downloads/ only removes the item from the webpage. 
Does not remove it from downloads.db so on browser restart download is still listed. 

Fix for 3 in #1108 

fixes: #1108


fictitiousexistence:
Added fix for 1 and 2 in #1180

PR #1120, opened 2025-09-05
unique_instance: allow opening command line URLs in the current tab

balejk:
For a kiosk use-case, it can be useful to be able to change the
currently open page externally. There are several options on how to
implement this but the simplest is to make use of the existing
unique_instance plugin and allow modifying its behaviour to open the
(single) URL provided by a secondary instance in the current tab.

Requested-by: puterboy <github@kosowsky.org>

---

@puterboy, please verify this updated version still works for you.


puterboy:
Thanks @balejk - this patch is extremely helpful for my 'kiosk' use case where I don't want to generate new tabs but rather keep everything in a single (full-screen) window. Without this new tabs are created that *obscure* old tabs and increase memory and CPU usage without even being visible or easily accessible to the user.
So this patch is critical to me!
Thanks!

puterboy:
Also as discussed, one could alternatively just add:
**ALTERNATIVE # 2**
```
        if _M.open_link_in_current_tab then
            w:navigate(u[1])
            table.remove(u, 1)
        end
```
before the existing code:
```
        if not _M.open_links_in_new_window then
            for _, uri in ipairs(u) do
                w:new_tab(uri)
            end
        end
```

This would (slightly) generalize the behavior in the PR so that if you send more than one url, the first is launched in the current active tab while the others generate new tabs.
(I also tested this alternative code and it works)

The argument for this approach is that  in some ways it is both:
1. Less disruptive to the existing code in that it only changes the processing of the first url
2. It doesn't in some ways "surprise" the user in that adding url's changes the behavior of the first url.

Another alternative interpretation for multiple tabs (that I haven't coded or tested) would be the following logic:
**ALTERNATIVE # 3**
1. First url replaces active tab
3. Subsequent url's replace other (non-active) tab perhaps in order that they were created 
4. After run out of tabs to replace, new tabs are created for remaining url's

The argument of this approach is that naturally extends the notion of reusing existing tabs as much as possible before creating new ones.

Of course, the argument for the original PR is that it minimally extends the current behavior to satisfy the primary use case request while additional more extensive use cases are currently not being requested.

There are pros/cons to each of these approaches of course -- though for most people (including me), the most important point is the single URL case which is covered by all 3 variants. While I personally prefer Alternative # 2, they all work for me.

c0dev0id:
I didn't read any discussions outside this PR.
If we're talking about the default luakit behavior, there should be no implicit logic like calling it with one URL does something else than calling it with multiple URLs. 

Do I understand it correctly that luakit should only ever show one tab?

What about an option `settings["window.startup_uri_mode"] = "newtab"`  -- or `"replace"`? That either adds URLs as tabs to the existing tab list (existing behavior), or simply replace all tabs with the list of URLs (reuse existing, then close surplus tabs).

Alternatively, we stick to the classic luakit customization approach and define the the `open_startup_uris` function before `userconf.lua` is loaded.

```lua
  window.open_startup_uris = function(uris, w)
      if w then
          for i, uri in ipairs(uris) do
              w:new_tab(uri, { switch = i == 1 })
          end
      else
          window.new(uris)
      end
  end

  require "userconf"
  ```
  
Then the user can redefine it within `userconf.lua`.

Then we call it.

  ```lua
  local w = (not luakit.nounique) and (session and session.restore())
  window.open_startup_uris(uris, w)
  ```

balejk:
> If we're talking about the default luakit behavior, there should be no implicit logic like calling it with one URL does something else than calling it with multiple URLs. 

The default is preserved by this patch.

> Do I understand it correctly that luakit should only ever show one tab?

I think not. The request was simply to allow replacing the currently
open tab programatically, for a kiosk use case.

> What about an option `settings["window.startup_uri_mode"] = "newtab"`  -- or `"replace"`? That either adds URLs as tabs to the existing tab list (existing behavior), or simply replace all tabs with the list of URLs (reuse existing, then close surplus tabs).

One of the discussed variants was replacing as much tabs as possible
(based on the number of links provided on the command line), as you
write, but leaving the others. This however seemed somewhat arbitrary
and the approach taken here is a more minimal to achieve what the
presented use case required. But what you write, i. e. closing the
redundant tabs, seems less arbitrary to me and may be sensible too.

> Alternatively, we stick to the classic luakit customization approach and define the the `open_startup_uris` function before `userconf.lua` is loaded.

I think I like this the best -- it's the most robust approach and means
less code to maintain on the luakit side. I can rework to PR to do that
if we agree on it.

puterboy:
I'm happy with whatever works best - kindly just let me know what I need to change if you use a different method as I currently am using @balejk's patch in a semi-popular Addon that I support for HomeAssistant that allows one to run a kiosk on the HAOS instance itself.

PR #1127, opened 2025-12-28
Fix pdfjs knob

msdemlei:
This fixes the pdf.js-disabling setting (upstream has changed the name of the internal feature).

### DIFF
--- a/widgets/webview.c
+++ b/widgets/webview.c
@@ -770,7 +770,11 @@ luaH_webview_set_pdfjs(lua_State *L)
         WebKitFeature *feature = webkit_feature_list_get(features, i);
         if (!strcmp(
                 webkit_feature_get_identifier(feature),
+#if WEBKIT_CHECK_VERSION(2, 43, 1)
+                "PDFJSViewer")) {
+#else
                 "PdfJSViewer")) {
+#endif
             webkit_settings_set_feature_enabled(settings, feature, enabled);
             break;
         }


balejk:
Seems to work now, thanks! I experimented with this setting in the past and was confused since it didn't seem to have any effect which was unfortunate because I prefer using the system viewer too.

You write the name has *changed* upstream, should this perhaps be under some WebKit version #if then?

msdemlei:
On Tue, Jan 20, 2026 at 05:29:50AM -0800, balejk wrote:
> balejk left a comment (luakit/luakit#1127)
>
> Seems to work now, thanks! I experimented with this setting in the past and was confused since it didn't seem to have any effect which was unfortunate because I prefer using the system viewer too.
>
> You write the name has *changed* upstream, should this perhaps be
> under some WebKit version #if then?

I think I'm too lazy to research when the change happened under the
rationale that whoever uses a webkit that's that old is in trouble
anyway?

On the other hand, if you have a checkout of the webkit source and
can run a quick git blame, I'd not oppose some preprocessor magic.

fictitiousexistence:
https://github.com/WebKit/WebKit/commit/395c75703661653f2a9178f311f92151416899e2
_"Rename PDFJSViewerEnabled setting to use all-caps PDF now that I fixed the generator."_

Looks like webkit 2.43.1 released 11/17/2023

I haven't seen anything in luakit that says how old of webkit versions it tends to keep compatibility with. Would be nice to have and clean up some of the old #ifs

msdemlei:
On Sun, Jan 25, 2026 at 11:05:07AM -0800, fictitiousexistence wrote:
> fictitiousexistence left a comment (luakit/luakit#1127)
>
> https://github.com/WebKit/WebKit/commit/395c75703661653f2a9178f311f92151416899e2
> _"Rename PDFJSViewerEnabled setting to use all-caps PDF now that I fixed the generator."_
>
> Looks like webkit 2.43.1 released 11/17/2023

Count me as unconvinced that we ought to worry about webkits that
old, but I've added a version guard and forced-pushed.

It'd be lovely if this could get merged, and a roll-up release from
which I could update the Debian package would also be great (because
then people using the package could again turn pdf.js off:-).

> I haven't seen anything in luakit that says how old of webkit
> versions it tends to keep compatibility with. Would be nice to have
> and clean up some of the old #ifs

I'd say that's for us to decide (where perhaps "fails completely" is
a different category from "soft failure" as in this case, where the
switch just doesn't do anything).

If I find . -name "*.c" | xargs grep "#if", however, I have to say
it's better than I had expected.  There's just 9 cases of conditional
compilation depending on webkit versions.

The limiting versions are 2.16.0, 2.18.0, 2.24.0, 2.26.0, and now
2.43.1.  Perhaps I'd drop the 2.1x switches?

But then the elephant in the room is GTK version checks.  I've not
actually checked what they do, but there's almost 40 of them, and
most of them in the 3.1x range.  Phewy.  Caring about old GTKs to me
is a different thing than caring about old Webkits (there's by and
large nothing wrong with keeping old libgtks around), so I don't
think there's much to be done about those.  Ah well.

PR #1139, opened 2026-05-19
feat: gtk4 and webkitgtk6.0 migration

616b2f:


c0dev0id:
You're on the right path here. Good start!

616b2f:
> You're on the right path here. Good start!

Thanks, in the last commit I used AI to help me migrate, I looked over it and it seems okay but I am not experienced enough in that area to say for sure. Would appreciate a look from you if you have time, especially around the JSCContext areas.

PluMGMK:
I tried this out for a while, and I noticed a few issues:
* Downloads don't seem to work at all (but uploads do)
* Following seems more likely to stop working on sites where it normally works (e.g. https://raymanpc.com) but I haven't dug into why
* Probably something to do with dark mode configuration: sometimes Irish Times articles (e.g. https://www.irishtimes.com/world/us/2026/07/08/ice-agent-fatally-shoots-motorist-during-road-stop-in-houston/) render with a black instead of white background, so the text is black-on-black and hence illegible.
* EDIT: Forgot to mention: tab order can get messed up too. Pressing `u` to restore a tab inserts it a position below where it should be, and you have to move it back and forth a few times with `<`/`>` or it doesn't get selected in the right order when using `J`/`K` to scroll through them.

616b2f:
> I tried this out for a while, and I noticed a few issues:
> * Downloads don't seem to work at all (but uploads do)
> * Following seems more likely to stop working on sites where it normally works (e.g. https://raymanpc.com) but I haven't dug into why
> * Probably something to do with dark mode configuration: sometimes Irish Times articles (e.g. https://www.irishtimes.com/world/us/2026/07/08/ice-agent-fatally-shoots-motorist-during-road-stop-in-houston/) render with a black instead of white background, so the text is black-on-black and hence illegible.
> * EDIT: Forgot to mention: tab order can get messed up too. Pressing `u` to restore a tab inserts it a position below where it should be, and you have to move it back and forth a few times with `<`/`>` or it doesn't get selected in the right order when using `J`/`K` to scroll through them.

Great thank you for the feedback, I will look into it, can you remember which commit exactly you used for testing or was it the recent one?

PluMGMK:
I tried two different ones. Unfortunately GitHub just says things like "2 days ago" beside commits which makes it hard to figure out which ones I was using at any given time. I'd love to meet the person who thought those kinds of datestamps were a good idea, they're everywhere on the internet and really annoying :unamused: 

I think the last one I used was indeed the latest one here, https://github.com/luakit/luakit/pull/1139/commits/ffa2d2b628ec0245dd4f8de08d77ec48d2ccee85

I also applied the first two commits from my PR #1148 over it.

PluMGMK:
Aha, that explains why YouTube embeds broke following in the original version! (the hints would appear but nothing would happen when you selected one)

616b2f:
> Aha, that explains why YouTube embeds broke following in the original version! (the hints would appear but nothing would happen when you selected one)

Yeah, I will try to drive this version for a while and hope that this was the last part of fixes and it's useable for daily use.

PS: pls try the version without your fixed on top, they should not be necessary in this version, I hope at least

PluMGMK:
What's interesting is that the GTK3 version has the problem I described, while my PR also somehow fixes it for the case of YouTube embeds on https://raymanpc.com (but not some other cases) – I don't really understand why… :confused:

nilFinx:
Is it possible to make this optional?

616b2f:
> Is it possible to make this optional?

What exactly?

nilFinx:
> > Is it possible to make this optional?
> 
> What exactly?

The GTK3 > GTK4 bump. I can see people wanting to keep it on GTK3.

616b2f:
> > > Is it possible to make this optional?
> > 
> > What exactly?
> 
> The GTK3 > GTK4 bump. I can see people wanting to keep it on GTK3.

Not for me to decide, but I don't think that we can afford to maintain both of them, because internally they're where a lot of changes for modern webkit and Wayland focused gtk4, this would be a huge afford to do future features for both.

616b2f:
FYI: I created a copr repo (https://copr.fedorainfracloud.org/coprs/616b2f/luakit/) with `luakit-nightly` to follow this PRs changes, if you use fedora based OS and like to test it and help me finish this feature.

PR #1146, opened 2026-06-02
Block tracking pings via lib/hyperlink_auditing_wm

fictitiousexistence:
As of WebKitGTK 2.50, the hyperlink-auditing setting in deprecated and has been removed. Pings are now sent by default.

This PR adds lib/hyperlink_auditing_wm to block all pings for Webkit 2.50+.


fictitiousexistence:
Note- we could also remove the hyperlink_auditing setting completely and the Webkit version checks and block pings via lua for all Webkit versions.

Issue #1119, opened 2025-08-25
Browser specific navigation doesn't work on YouTube


jasonfuchs:
**Current Behavior:**
Can't use keybinds to navigate (follow mode, j,k don't work). Scrolling with the mouse wheel and arrow keys (in insert mode) works.

It prints the following 6 errors:

```
** (WebKitWebProcess:189537): CRITICAL **: 16:44:54.822: WebKitDOMDocument *webkit_web_page_get_dom_document(WebKitWebPage *): assertion 'WEBKIT_IS_WEB_PAGE(webPage)' failed

** (WebKitWebProcess:189537): CRITICAL **: 16:44:54.822: WebKitDOMDOMWindow *webkit_dom_document_get_default_view(WebKitDOMDocument *): assertion 'WEBKIT_DOM_IS_DOCUMENT(self)' failed

** (WebKitWebProcess:189537): CRITICAL **: 16:44:54.822: void webkit_dom_dom_window_scroll_to(WebKitDOMDOMWindow *, gdouble, gdouble): assertion 'WEBKIT_DOM_IS_DOM_WINDOW(self)' failed

** (WebKitWebProcess:189537): CRITICAL **: 16:44:54.822: glong webkit_dom_dom_window_get_scroll_x(WebKitDOMDOMWindow *): assertion 'WEBKIT_DOM_IS_DOM_WINDOW(self)' failed

** (WebKitWebProcess:189537): CRITICAL **: 16:44:54.822: glong webkit_dom_dom_window_get_scroll_y(WebKitDOMDOMWindow *): assertion 'WEBKIT_DOM_IS_DOM_WINDOW(self)' failed

** (WebKitWebProcess:189537): CRITICAL **: 16:44:54.822: guint64 webkit_web_page_get_id(WebKitWebPage *): assertion 'WEBKIT_IS_WEB_PAGE(webPage)' failed
```

**Desired Behavior:**
Make follow mode etc. usable.

**How can we reproduce it (step by step):**
1. install `luakit`, `gts-plugins-good`, `gts-plugins-bad` and `gts-plugins-ugly` on archlinux
2. visit youtube.com
3. use j,k/follow mode

**Environment:**

Linux Distribution & Version: Arch Linux
Output of `luakit --version`: luakit 2.4.0
  built with: webkit 2.46.6 (installed version: 2.48.5)
                 GTK 3.24.49 
                GLIB 2.84.0 
                SOUP 3.6.4

**Note about webkit issues:**

[...]

PR #1148, opened 2026-07-05
Fix IPC endpoints desyncing from WebView widgets, and associated Undefined Behaviour (coredumps)

PluMGMK:
I did a bit of an audit on how refcounting is managed for IPC endpoints, and explained what I found starting here: https://github.com/luakit/luakit/issues/1067#issuecomment-4882449713

While this doesn't solve the underlying data race (https://github.com/luakit/luakit/commit/525c03d19d99aabcc80d40f09b1a78b234093449#diff-23e66972634a71b55ae442fbb3cb8ce1bda6e4f0aa885e471affbf4e28b071c6R34 may do that), this does eliminate a whole suite of use-after-free errors (which could also occur when simply closing the browser) and possibly also some memory leaks.

This should finally put an end to #1067, though again the data race needs to be fixed to prevent some of the co-occurring annoyances.

### DIFF
--- a/common/ipc.c
+++ b/common/ipc.c
@@ -274,6 +274,7 @@ ipc_endpoint_decref(ipc_endpoint_t *ipc)
     }
     ipc->status = IPC_ENDPOINT_FREED;
     g_slice_free(ipc_endpoint_t, ipc);
+    debug("Freeing IPC endpoint 0x%lx", (unsigned long)ipc);
 }
 
 void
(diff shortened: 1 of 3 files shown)


PluMGMK:
I've added a hacky but simple and effective solution for the desyncing of the `page-id`s in the latest commit. So if this is pulled in its current form, it should resolve all aspects of #1067, and #1119.

